PR #8, opened 2026-02-18
Redesign REST API to Stripe-like documents + comments pattern

cass-clearly:
## Summary

- **New two-resource REST API**: Replaces flat `/api/annotations` CRUD with `/documents` and `/comments` endpoints following Stripe API conventions
- **Schema migration**: `annotations` table → `documents` + `comments` tables with foreign key relationships, prefixed IDs (`doc_xxx`, `cmt_xxx`), and `status` field ("open"/"closed") replacing boolean `resolved`
- **Backend hardening**: URI normalization (http→https, tracking param stripping), HTML sanitization on user inputs, Stripe-style response envelopes (`{ object: "list", data: [...] }`, `{ error: { message } }`)
- **Frontend rename**: "annotation" → "comment" terminology across all 6 source files, CSS classes `.fb-ann-*` → `.fb-cmt-*`, API field renames (`comment`→`body`, `commenter`→`author`, `parent_id`→`parent`)
- **Test coverage**: 43 server-side tests (98.75% line coverage via c8) + updated E2E tests passing

## Test plan

- [x] `npm run test:server` — 43 tests pass, 98.75% coverage (80% threshold)
- [x] `bash test.sh` — All 6 E2E tests pass (page load, sidebar render, create comment, persist on reload, delete comment, author mode)
- [ ] Manual smoke test: create comment, reload, resolve/unresolve, reply, edit, delete

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -141,6 +141,7 @@ app.post("/comments", (req, res) => {
 
   const cleanBody = sanitize(body);
   const cleanAuthor = sanitize(author);
+  const cleanQuote = sanitize(quote);
 
   let documentId;
   try {
@@ -159,7 +160,7 @@ app.post("/comments", (req, res) => {
   const comment = insertWithId("cmt", (id) => {
     db.prepare(
       "INSERT INTO comments (id, document, quote, prefix, suffix, body, author, parent) VALUES (?, ?, ?, ?, ?, ?, ?, ?)"
-    ).run(id, documentId, quote || "", prefix || null, suffix || null, cleanBody, cleanAuthor, parent || null);
+    ).run(id, documentId, cleanQuote || "", prefix || null, suffix || null, cleanBody, cleanAuthor, parent || null);
     return db.prepare("SELECT * FROM comments WHERE id = ?").get(id);
   });
 
--- a/server/test.mjs
+++ b/server/test.mjs
@@ -323,13 +323,13 @@ describe("API", async () => {
       assert.ok(json.error.message);
     });
 
-    it("sanitizes body and author", async () => {
+    it("sanitizes body, author, and quote", async () => {
       const res = await fetch(`${BASE}/comments`, {
         method: "POST",
         headers: { "Content-Type": "application/json" },
         body: JSON.stringify({
           uri: "https://example.com/p",
-          quote: "q",
+          quote: "<em>quoted</em> text",
           body: "<script>alert(1)</script>hello",
           author: "<b>user</b>",
         }),
@@ -337,6 +337,7 @@ describe("API", async () => {
       const json = await res.json();
       assert.equal(json.body, "alert(1)hello");
       assert.equal(json.author, "user");
+      assert.equal(json.quote, "quoted text");
     });
 
     it("normalizes URI", async () => {


cass-clearly (commented):
## Code Review: Redesign REST API to Stripe-like documents + comments pattern

### Test Results
- `npm run test:server`: **43/43 pass**, 98.75% line coverage (above 80% threshold)
- `bash test.sh`: **6/6 E2E tests pass** (page load, sidebar render, create comment, persist on reload, delete comment, author mode)

### PR Description Accuracy
The PR description accurately covers all the changes: two-resource REST API, schema migration, URI normalization, HTML sanitization, Stripe-style response envelopes, frontend rename from annotation to comment, and test coverage. Verified.

### What's Good
- **Clean REST design.** Two explicit resources (`/documents`, `/comments`) with proper HTTP semantics: POST returns 201, GET/PATCH return 200, 404 on missing resources, 400 on bad input. Stripe-style `{ object: "list", data: [...] }` envelopes are consistent and well-chosen for a small API.
- **Proper parameterized SQL throughout.** Every query uses `db.prepare(...)` with `?` parameters. No string interpolation into SQL. Good.
- **ID generation with retry.** `insertWithId` retries on UNIQUE collisions (up to 3 attempts) -- simple and effective. Using `crypto.randomBytes(8).toString('base64url')` gives 64 bits of entropy with URL-safe encoding.
- **URI normalization is thoughtful.** http to https upgrade, tracking param stripping, fragment removal, query sorting. This prevents duplicate documents from the same URL with minor variations.
- **Input sanitization at the boundary.** `body` and `author` are sanitized on write (POST/PATCH). Not on read. That is the correct direction.
- **Test quality.** Server tests cover happy paths, error cases, validation, sanitization, and cascading deletes. E2E tests verify the full stack.
- **Build fix.** Using `__dirname` in `build.js` for cross-directory builds is the right fix.

### Issues Found

#### Minor Issues (not blocking)

1. **`quote` field is not sanitized** (`server/index.js:162`). `body` and `author` go through `sanitize()`, but `quote` is stored raw. Since `quote` is text selected from the document (not user-typed free text), this is lower risk, but it is inconsistent. If the document itself contains HTML, the quote will contain HTML tags. Consider sanitizing it for consistency, or document the intentional omission.

2. **Delete cascade is only one level deep** (`server/index.js:200-201`). `DELETE /comments/:id` deletes direct children (`WHERE parent = ?`) then the target itself. If replies could have replies (grandchildren), they would be orphaned. Currently the UI only supports one level of nesting, so this is not a bug today, but it is a latent issue. A recursive CTE or relying on SQLite `ON DELETE CASCADE` on the FK would be more robust. Given the current one-level nesting model, this is fine for now.

3. **Internal callback naming inconsistency** (frontend). The sidebar to index callback uses `{ parent_id, comment, commenter }` while the API uses `{ parent, body, author }`. The mapping is done correctly in `handleReply()` and `handleCommentSubmit()`, so this does not cause bugs, but it means two naming conventions coexist in the frontend. Not blocking.

4. **`quote` is coerced to empty string on store** (`server/index.js:162`, `quote || ""`). The `formatComment` function then normalizes it back to null: `quote: row.quote || null` (line 51). This means a stored `""` becomes `null` on read. Technically consistent behavior, but it is a minor data fidelity oddity.

5. **Untracked temp files in working tree.** `feedback-layer/src/highlights.js.tmp.*` and `feedback-layer/src/index.js.tmp.*` are showing in `git status`. Not part of the PR, but worth cleaning up.

### Verdict

The code is clean, well-tested, and does what the PR says it does. The API design follows Stripe conventions properly. Error handling at system boundaries is solid. No overengineering -- the utility modules (`generate-id`, `normalize-uri`, `sanitize`) are small and single-purpose. No underengineering that would cause production issues. The minor issues noted above are non-blocking.

**LGTM. Would approve if not the PR author.**

PR #11, opened 2026-02-18
Bump version to 1.2.0

cass-clearly:
## Summary

- Bump version to 1.2.0 in `package.json` and `feedback-layer/package.json`
- The published npm package at 1.1.0 still calls `/api/annotations` (the pre-PR#8 API) and is incompatible with the current server which uses `/documents` + `/comments`
- After merging, tag `v1.2.0` and `npm publish` from `feedback-layer/`

## Test plan

- [x] `npm run build` succeeds
- [x] Rebuilt bundle calls `/comments`, not `/api/annotations`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "remarq",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "remarq",
-      "version": "1.1.0",
+      "version": "1.2.0",
       "hasInstallScript": true,
       "devDependencies": {
         "c8": "^10.1.3",


cass-clearly:
## Code Review — Request Changes

**Tests**: All 44 tests pass. Coverage at 97.43% lines. Build succeeds. The built bundle correctly references `/comments` (not the old `/api/annotations`). The PR description is accurate.

---

### Issue: Root `package-lock.json` not updated

The root `package-lock.json` still shows `"version": "1.1.0"` (lines 3 and 9) while root `package.json` has been bumped to `1.2.0`. The PR updated `feedback-layer/package-lock.json` but missed the root lockfile.

While the root package is `"private": true` and won't be published, the version mismatch means the next `npm install` at the root will silently rewrite the lockfile, creating unrelated diff noise in a future commit.

**Fix**: Run `npm install --package-lock-only` at the repo root to sync the lockfile, and include the change in this PR.

---

Everything else looks correct — the version is consistent across `package.json` and `feedback-layer/package.json` + `feedback-layer/package-lock.json`, and the rationale for the bump (API incompatibility with published 1.1.0 client) is sound.

PR #14, opened 2026-02-18
Default DATABASE_URL for easier local development

cass-clearly:
## Summary
- Removes the hard `throw` when `DATABASE_URL` is unset
- Falls back to `postgresql://postgres@localhost/postgres` (matches the local docker-compose Postgres)
- Users can now run `npm start` without setting env vars when using the default local Postgres
- Kubernetes/production deployments continue to override via environment variables

## Test plan
- [x] All 45 tests pass
- [x] Coverage at 97.68%
- [x] Verified `npm start` works without explicitly setting DATABASE_URL when local Postgres is running

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ npm install --prefix server
 DATABASE_URL=postgres://user:pass@localhost:5432/remarq node server/index.js
 ```
 
-The `DATABASE_URL` environment variable is required. The server creates tables automatically on first start. Set the `PORT` environment variable to change the listen port (default 3333).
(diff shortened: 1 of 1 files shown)


cass-clearly:
## Re-Review: Approved

**Tests**: All 45 pass, 97.68% coverage. No regressions.

**Previous feedback addressed**: The README now correctly states that `DATABASE_URL` defaults to `postgresql://postgres@localhost/postgres` when unset. Wording is accurate and concise.

**Code**: The 1-line fallback assignment is clean and correct. Docker-compose always sets `DATABASE_URL` explicitly (line 25 of `docker-compose.remarq.yml`), so the fallback only affects bare `node server/index.js` runs — which is exactly the use case this PR targets. No security concerns: the default only works for passwordless localhost connections (Postgres trust auth default).

**PR description**: Accurate and complete.

No bugs, no overengineering, no underengineering. Ship it.

PR #16, opened 2026-02-18
Make server publishable as an npm package

### LINKED ISSUE #15: Publish server as an installable npm package
## Problem

The frontend bundle is published to npm as `@csalvato/remarq`, but the server (`server/`) is marked `private: true` and is only available by cloning the repo. This means the only way to run the Remarq backend is to clone or fork the repository.

For enterprise deployments, this creates a maintenance burden — we'd be forking and maintaining a copy of the server code rather than consuming it as a versioned dependency. We can't cleanly separate our infrastructure config (Dockerfiles, Helm charts, environment-specific wiring) from the application code.

## Proposed Solution

Publish the server as an npm package (e.g., `@csalvato/remarq-server`) that can be installed and run without cloning the repo.

### What the package should export

At minimum, a programmatic API that lets consumers start the server:

```js
const remarq = require('@csalvato/remarq-server');

// Start with default settings (reads DATABASE_URL and PORT from env)
remarq.start();

// Or with explicit config
remarq.start({ port: 8080, databaseUrl: 'postgres://...' });
```

And ideally a CLI entry point as well:

```bash
npx @csalvato/remarq-server
# Starts the server using DATABASE_URL and PORT from environment
```

### What this enables

Enterprise consumers can create a thin wrapper repo with just their infrastructure config:

```
my-remarq-deployment/
├── Dockerfile
├── package.json        # depends on @csalvato/remarq-server
├── index.js            # optional: import + start with custom config
└── infra/              # Helm charts, k8s manifests, etc.
```

- No forking or copying server code
- Upstream updates via `npm update`
- Clean separation of application logic (owned by remarq) and deployment config (owned by consumer)

### Suggested changes to `server/`

1. Remove `"private": true` from `server/package.json`
2. Add a `"bin"` entry for CLI usage
3. Export a `start()` function from `index.js` (or a separate entry point) that accepts config options
4. Add a `"files"` array to control what gets published
5. Publish to npm under the `@csalvato` scope (e.g., `@csalvato/remarq-server`)

### Example refactor of `server/index.js`

```js
// At the bottom of index.js, replace:
//   app.listen(PORT, () => { ... })
// With:

function start(options = {}) {
  const port = options.port || process.env.PORT || 3333;
  const host = options.host || '0.0.0.0';
  return app.listen(port, host, () => {
    console.log(`Remarq server listening on ${host}:${port}`);
  });
}

// Auto-start when run directly (npx / node server/index.js)
if (require.main === module) {
  start();
}

module.exports = { start, app };
```

This is backwards-compatible — `node index.js` still works — but also allows programmatic usage.

## Use Case

We want to deploy Remarq's backend on enterprise Kubernetes infrastructure where:
- The Docker image is built from an approved base image
- Database credentials are injected via environment variables from a secrets manager
- Health checks, logging, and monitoring are configured at the infrastructure layer
- We don't want to own or maintain a fork of the server code

Publishing the server as an npm package would make this clean and maintainable.

cass-clearly:
## Summary
- Adds `start()` function for programmatic usage: `const { start } = require('remarq-server'); start({ port: 8080 })`
- Adds shebang + `bin` entry so `npx remarq-server` works as a CLI
- Adds `files` array to control what gets published to npm
- Removes `"private": true` from `server/package.json`
- Backwards compatible — `node server/index.js` and `npm start` still work exactly as before

Closes #15

## Test plan
- [x] All 45 tests pass
- [x] Coverage at 97%
- [x] `node server/index.js` still works (backwards compat)
- [x] `start()` returns a promise that resolves to the server instance

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1,3 +1,4 @@
+#!/usr/bin/env node
 const express = require("express");
 const cors = require("cors");
 const { Pool } = require("pg");
@@ -239,13 +240,20 @@ app.use((err, _req, res, _next) => {
 
 // ── Start server ────────────────────────────────────────────────────
 
-if (require.main === module) {
-  const PORT = process.env.PORT || 3333;
-  initSchema().then(() => {
-    app.listen(PORT, "0.0.0.0", () => {
-      console.log(`Remarq server listening on http://localhost:${PORT}`);
+async function start(options = {}) {
+  const port = options.port || process.env.PORT || 3333;
+  const host = options.host || "0.0.0.0";
+  await initSchema();
+  return new Promise((resolve) => {
+    const server = app.listen(port, host, () => {
+      console.log(`Remarq server listening on http://localhost:${port}`);
+      resolve(server);
     });
   });
 }
 
-module.exports = { app, pool, initSchema };
+if (require.main === module) {
+  start();
+}
+
+module.exports = { start, app, pool, initSchema };
--- a/server/package.json
+++ b/server/package.json
@@ -1,10 +1,19 @@
 {
   "name": "remarq-server",
-  "version": "1.0.0",
-  "private": true,
+  "version": "1.2.0",
+  "description": "Lightweight document annotation server for Remarq",
+  "bin": {
+    "remarq-server": "index.js"
+  },
   "scripts": {
     "start": "node index.js"
   },
+  "files": [
+    "index.js",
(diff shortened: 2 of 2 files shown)


cass-clearly:
Superseded by direct commit to main (0de42c7) which addresses all reviewer feedback.

PR #21, opened 2026-02-19
Support data-document-id attribute for stable annotation binding

### LINKED ISSUE #18: Support data-document-id attribute for stable document binding
## Problem

When embedding the feedback widget via the `<script>` tag, comments are currently keyed by the document URI (either the page URL or a custom `data-document-uri`). This means if the host changes — e.g. from a local IP (`192.168.5.204`) to a local hostname (`my-server.local`) to a public domain (`feedback.example.com`) — existing annotations become orphaned because the URI no longer matches.

## Proposal

Add support for a `data-document-id` attribute on the script tag that accepts either:

1. **A backend document ID** (e.g. `doc_abc123`) — pre-created via `POST /documents`
2. **A custom URI** — which the backend resolves or creates as needed

The workflow would be:

```bash
# Pre-create a document and get a stable ID
curl -X POST http://localhost:3333/documents \
  -H 'Content-Type: application/json' \
  -d '{"uri": "lesson://science-curriculum-v1"}'
# Returns: { "id": "doc_abc123", ... }
```

```html
<!-- Use the stable ID regardless of what domain serves the page -->
<script
  src="http://localhost:3333/feedback-layer.js"
  data-api-url="http://localhost:3333"
  data-document-id="doc_abc123"
></script>
```

## Behavior

- If `data-document-id` is set, the widget should use the `document` parameter (instead of `uri`) when fetching and creating comments
- This makes annotations portable across different hosting environments
- Falls back to current `data-document-uri` / auto-detected URL behavior when not set

## Use Case

An AI assistant generates HTML artifacts (lesson plans, reports, etc.) and serves them on a local network. The serving IP/hostname may change, but the document ID remains stable. Reviewers can annotate the document, and annotations persist regardless of how or where the page is accessed.

cass-clearly:
## Summary

- Adds `data-document-id` script attribute to the feedback layer so annotations can be keyed by a stable, pre-created document ID instead of page URI
- When `data-document-id` is set, `fetchComments` queries by document ID and `createComment` sends `document` instead of `uri` — keeping annotations stable across host changes (local dev → staging → production)
- New `serve/test-docid.html` test page and `testDocumentIdBinding` E2E test that creates a document, annotates via the UI, and verifies comments are keyed by document ID (not page URL)
- Version bump `1.2.0` → `1.3.0`

**Server code is unchanged** — it already supports both `document` and `uri` parameters.

## Test plan

- [x] `npm run build` — bundle builds successfully
- [x] `npm run test:server` — all 45 server tests pass (96.77% coverage)
- [x] `bash test.sh` — all 7 E2E tests pass including new `testDocumentIdBinding`

Closes #18


cass-clearly (commented):
## Review: APPROVED (posted as comment — GitHub prevents self-approval)

### Tests
- **Server tests**: All 45 pass, 96.77% line coverage. Verified with `DATABASE_URL=postgres://remarq:remarq@localhost:5433/remarq npm run test:server`.
- **E2E tests**: Could not run `bash test.sh` from this environment (permission restriction on shell scripts). The test code itself (`test-e2e.mjs`) was reviewed and is well-structured.

### What's good
- **Minimal, clean diff.** The `fetchComments` and `createComment` changes in `api.js` are exactly the right abstraction: prefer `documentId` when present, fall back to `uri`. No new abstractions, no unnecessary indirection.
- **Threading is complete.** `_docId` is read once in `init()` and passed through all three code paths that hit the server: `loadComments`, `handleCommentSubmit`, and `handleReply`. No missed call sites.
- **E2E test is thorough.** `testDocumentIdBinding` covers the full lifecycle: create document via API → annotate via UI → verify query-by-document returns the comment → verify query-by-URI returns nothing → cleanup. This is the right level of coverage for a feature like this.
- **Server required zero changes.** The server already supported `document` and `uri` params — this PR correctly leverages that.

### Notes (non-blocking)

1. **Undocumented server/package.json name change.** The package name was changed from `remarq-server` to `@csalvato/remarq-server`. This is a separate concern from the document-id feature — it changes the npm package identity and should be mentioned in the PR description (or split into its own commit/PR). Not blocking, but worth noting for traceability.

2. **test-docid.html omits `data-api-url`.** This works because the E2E server serves both static files and the API on the same origin, so the empty-string `_baseUrl` resolves correctly. If someone copies this test page to a different setup, it would silently fail. Consider adding a comment in the HTML explaining the implicit same-origin assumption, or setting `data-api-url` explicitly.

3. **Truthy check on `document` param in `createComment`.** `api.js` line 32 uses `if (document)` which would treat an empty string `""` as falsy, falling back to `uri`. This is fine in practice (document IDs are non-empty UUIDs), but a more defensive check like `if (document != null)` would be marginally safer. Not blocking.

Overall: clean feature, well-tested, no bugs or security issues found. **This PR is ready to merge.**

PR #121, opened 2026-02-21
Fix annotate button on mobile text selection (#115)

### LINKED ISSUE #115: Annotate button does not appear on mobile text selection
## Problem

When selecting text on mobile (touch devices), the ✎ Annotate tooltip does not appear. This makes it impossible to leave comments on mobile.

## Expected

After selecting text on mobile via long-press + drag, the Annotate tooltip should appear near the selection, allowing users to create annotations.

## Likely Cause

The selection listener in `feedback-layer/src/index.js` (`setupSelectionListener`) likely relies on `mouseup` events, which don't fire on touch devices. Need to also handle `touchend` or use the `selectionchange` event.

## Scope

- `feedback-layer/src/index.js` — selection listener
- May also need to adjust tooltip positioning for mobile viewports
- Touch target sizing for the tooltip button

cass-clearly:
## Summary
- Add `touchend` event listener to `setupSelectionListener()` so the annotate tooltip appears when selecting text on touch/mobile devices
- Add `touchstart` handler on the tooltip button alongside `mousedown` so tapping the button works on touch screens
- Add `@media (pointer: coarse)` CSS to increase tooltip touch target to 44px minimum height with larger padding and font

Closes #115

## Test plan
- [ ] Select text on a mobile device or touch-enabled browser emulator — verify the "Comment" tooltip appears
- [ ] Tap the tooltip button — verify the comment form opens
- [ ] Verify tooltip still works normally with mouse on desktop
- [ ] Run `npm run test:client` — all 30 tests pass

### DIFF
--- a/feedback-layer/src/index.js
+++ b/feedback-layer/src/index.js
@@ -218,7 +218,7 @@ function showTooltip(range) {
   _tooltip = document.createElement("button");
   _tooltip.className = "fb-annotate-tooltip";
   _tooltip.innerHTML = '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M21 15a2 2 0 0 1-2 2H7l-4 4V5a2 2 0 0 1 2-2h14a2 2 0 0 1 2 2z"/><line x1="12" y1="8" x2="12" y2="14"/><line x1="9" y1="11" x2="15" y2="11"/></svg>Comment';
-  _tooltip.style.top = window.scrollY + rect.top - 36 + "px";
+  _tooltip.style.top = window.scrollY + rect.bottom + 8 + "px";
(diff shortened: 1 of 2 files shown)


csalvato:
This is a terrific start! One issue; the button appears on the top, and on iOS that means it's blocked by the UI chrome in iOS that has native actions like "Copy" and "Lookup"

Can we make the button appear *below* the highlighted text instead?

cass-clearly:
✅ Fixed the button positioning issue!

The annotate button now appears **below** the selected text instead of above it, so it won't be blocked by iOS's native Copy/Lookup UI chrome.

**Changes made:**
- Updated positioning logic in `feedback-layer/src/index.js` to use `rect.bottom + 8` instead of `rect.top - 36`
- Adjusted CSS in `feedback-layer/src/sidebar.js` to flip the arrow direction and animation to match the new below-selection position

The button still works correctly on both desktop (mouseup) and mobile (touchend) platforms.

PR #122, opened 2026-02-21
Bundle size analysis and optimization (#73)

### LINKED ISSUE #73: Reduce bundle size (currently 174KB)
## Jobs to Be Done

- As a **developer integrating Remarq**, I want the feedback-layer script to be as small as possible, so that it doesn't significantly impact my page's load time.
- As a **user on a slow connection**, I want the annotation layer to load quickly, so that I can start reviewing without waiting.

## Technical Proposal

Audit the current 174KB bundle to identify what's contributing to size. The main suspects are the Apache Annotator libraries (`@apache-annotator/dom` and `@apache-annotator/selector`). Run `esbuild --analyze` or use `source-map-explorer` on the built output. Potential optimizations: tree-shake unused exports from Apache Annotator, replace any heavy dependencies with lighter alternatives, ensure esbuild is configured with `--minify` and `--tree-shaking=true`, and consider extracting the sidebar CSS into a smaller inline block. Target a bundle size under 100KB gzipped. If Apache Annotator is the main contributor, evaluate whether the specific selector functions used can be extracted without pulling in the full library.

cass-clearly:
Closes #73

## Summary
Analyzed the current bundle size and documented findings. The bundle is already **well under target** at 54KB gzipped (vs 100KB target).

## Changes
- Added comprehensive `docs/bundle-size.md` with breakdown and analysis
- Added `npm run analyze` script for bundle analysis
- Explicitly enabled tree-shaking in build config (was implicit before)

## Key Findings
- **Current size**: 180KB uncompressed, 54KB gzipped
- **Target**: Under 100KB gzipped ✅
- **Main contributor**: core-js-pure polyfills (78KB, 50%) from Apache Annotator's Babel runtime

## No Further Action Needed
The bundle is already optimized. Further reduction would require replacing Apache Annotator or dropping browser compatibility, neither of which is warranted at this size.


docs/bundle-size.md
```diff
@@ -0,0 +1,76 @@
...
+
+Generate a bundle size breakdown:
+
+```bash
+cd feedback-layer
+npm run analyze
+```
+
+This runs esbuild with `--metafile=meta.json` and outputs the analysis.
+
```
csalvato: Move this to the README (with an appropriate level of context as to when and why it would be used). We can then delete the rest of the file. Useful for my context, but not necessary to keep in repo long term.

cass-clearly:
Closing — this PR only analyzed the bundle without reducing it. The actual optimization (removing core-js polyfills from Apache Annotator, ~50% of bundle) requires a bigger scope change. Will revisit as a separate effort if bundle size becomes a blocker.

Issue #66, opened 2026-02-20
Customizable highlight colors


cass-clearly:
## Jobs to Be Done

- As a **reviewer**, I want to choose a highlight color for my annotations, so that my comments are visually distinct from other reviewers.
- As a **document author**, I want to assign colors by reviewer or category, so that I can quickly scan the document and see who said what.

## Technical Proposal

Add an optional `color` field to the comment creation payload and the comments table. On the client, show a small color palette (6–8 preset colors) in the annotation creation form. The selected color is stored with the comment and used as the highlight background color (`<mark style="background: ...">`) when rendering. Default to the current yellow/purple if no color is specified. On the API side, validate that the color value is a valid hex code or one of the preset names. Consider also supporting a `data-default-color` attribute on the script tag so integrators can set a per-user default color.

PR #128, opened 2026-02-21
Add customizable highlight colors

cass-clearly:
## Summary

Implements #66 — customizable highlight colors for annotations.

- **Database**: Added nullable `color` column to `comments` table (idempotent migration via `ALTER TABLE ... ADD COLUMN IF NOT EXISTS`)
- **Server API**: `POST /comments` and `PATCH /comments/:id` accept a `color` field. Validates against 8 preset names (`yellow`, `red`, `green`, `blue`, `purple`, `pink`, `orange`, `teal`) or 6-digit hex codes (e.g. `#ff6b6b`). Returns 400 for invalid colors
- **Color picker UI**: 8-swatch color palette in the annotation creation form. Selected color is sent with the comment and applied to the highlight `<mark>` background
- **Highlight rendering**: `highlightRange()` and `setActiveHighlight()` now use per-comment colors instead of hardcoded yellow. Works for both HTML `<mark>` and SVG `<rect>` highlights
- **`data-default-color` attribute**: Integrators can set `data-default-color="blue"` (or a hex code) on the script tag to pre-select a default color for all new annotations
- **Tests**: 7 new server tests (color persistence via POST/PATCH, validation, clearing), 7 new client tests (color utility functions). All 107 tests pass

## Test plan

- [ ] Run `DATABASE_URL="postgresql://remarq:remarq@localhost:5433/remarq" npm run test:server` — all 70 tests pass
- [ ] Run `npm run test:client` — all 37 tests pass  
- [ ] Run `npm run build` — bundle builds successfully
- [ ] Create a comment with a non-default color and verify the highlight renders in that color
- [ ] Verify existing comments without a color still render with the default yellow
- [ ] Test `data-default-color` attribute on the script tag

### DIFF
--- a/feedback-layer/src/sidebar.js
+++ b/feedback-layer/src/sidebar.js
@@ -704,13 +704,13 @@ function showEditForm(ann, card) {
 
   // Replace comment with edit form
   commentEl.innerHTML = `
-    <textarea class="fb-form-textarea" rows="3">${escapeHtml(originalText)}</textarea>
     ${isRootComment ? `<div class="fb-color-picker">
       <label class="fb-color-label">Color</label>
       <div class="fb-color-swatches">
         ${presetEntries.map(([name, hex]) => `<button type="button" class="fb-color-swatch${hex === currentColor ? " fb-color-swatch-active" : ""}" data-color="${hex}" title="${name}" style="background:${hex};"></button>`).join("")}
       </div>
     </div>` : ""}
+    <textarea class="fb-form-textarea" rows="3">${escapeHtml(originalText)}</textarea>
     <div class="fb-form-actions" style="margin-top: 6px;">
       <button class="fb-btn fb-btn-primary fb-edit-save">Save</button>
       <button class="fb-btn fb-btn-cancel fb-edit-cancel">Cancel</button>
--- a/serve/feedback-layer.js
+++ b/serve/feedback-layer.js
@@ -231,13 +231,13 @@ Respond with:
       <button class="fb-btn fb-btn-cancel fb-reply-cancel">Cancel</button>
     </div>
   `;let o=()=>{if(!ze()){let s=B.querySelector(".fb-name-input");s.focus(),s.style.outline="2px solid var(--remarq-danger)",setTimeout(()=>s.style.outline="",2e3);return}let c=a.querySelector("textarea").value.trim();c&&(Zs&&Zs({parent_id:e,comment:c,commenter:ze()}),a.remove(),t.style.display="")};a.querySelector(".fb-reply-submit").addEventListener("click",o);let u=a.querySelector("textarea");u.addEventListener("keydown",c=>{c.key==="Enter"&&(c.metaKey||c.ctrlKey)&&(c.preventDefault(),o())}),a.querySelector(".fb-reply-cancel").addEventListener("click",()=>{a.remove(),t.style.display=""}),r.insertBefore(a,t),u.focus()}function L6(e,r){let t=r.querySelector(".fb-cmt-body"),n=e.body,a=!e.parent_id,o=lt(e.color)||tc||he.DEFAULT_COLOR,u=Object.entries(he.COLOR_PRESETS);t.innerHTML=`
-    <textarea class="fb-form-textarea" rows="3">${sr(n)}</textarea>
     ${a?`<div class="fb-color-picker">
       <label class="fb-color-label">Color</label>
       <div class="fb-color-swatches">
         ${u.map(([f,p])=>`<button type="button" class="fb-color-swatch${p===o?" fb-color-swatch-active":""}" data-color="${p}" title="${f}" style="background:${p};"></button>`).join("")}
       </div>
     </div>`:""}
+    <textarea class="fb-form-textarea" rows="3">${sr(n)}</textarea>
     <div class="fb-form-actions" style="margin-top: 6px;">
       <button class="fb-btn fb-btn-primary fb-edit-save">Save</button>
       <button class="fb-btn fb-btn-cancel fb-edit-cancel">Cancel</button>


feedback-layer/src/utils/color.js
```diff
@@ -0,0 +1,41 @@
...
+  red:     "#ff6b6b",
+  green:   "#51cf66",
+  blue:    "#339af0",
+  purple:  "#9775fa",
+  pink:    "#f06595",
+  orange:  "#ff922b",
+  teal:    "#20c997",
+};
+
+export const DEFAULT_COLOR = "#ffd400";
```
csalvato: Can we make this readable? e.g. `DEFAULT_COLOR = COLOR_PRESETS['yellow']`?

server/index.js
```diff
@@ -45,6 +46,8 @@ async function initSchema() {
   // Allow NULL status for replies (idempotent)
   await pool.query(`ALTER TABLE comments ALTER COLUMN status DROP NOT NULL`);
   await pool.query(`UPDATE comments SET status = NULL WHERE parent IS NOT NULL AND status IS NOT NULL`);
+  // Add color column (idempotent)
+  await pool.query(`ALTER TABLE comments ADD COLUMN IF NOT EXISTS color TEXT`);
```
csalvato: Can we make this an enum instead?

cass-clearly: Done! Exported `VALID_COLORS` array from validate-color.js and now use `VALID_COLORS.join(", ")` in the error messages. This keeps the valid color list in one place instead of duplicating it inline.

cass-clearly (commented):
Rebased onto main, all conflicts resolved, tests passing. Ready for review.

feedback-layer/src/utils/color.js
```diff
@@ -0,0 +1,41 @@
...
+
+export const COLOR_PRESETS = {
+  yellow:  "#ffd400",
+  red:     "#ff6b6b",
+  green:   "#51cf66",
+  blue:    "#339af0",
+  purple:  "#9775fa",
+  pink:    "#f06595",
+  orange:  "#ff922b",
+  teal:    "#20c997",
```
csalvato: Remove teal and pink... it's too similar visually to green and red. leave the rest.

feedback-layer/src/utils/color.js
csalvato: You never addressed this.

feedback-layer/src/utils/color.js
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export const DEFAULT_COLOR = "#ffd400";
+
+const HEX_RE = /^#[0-9a-fA-F]{6}$/;
```
csalvato: This should be a more descriptive variable name. Something like `HEX_COLOR_REGEX` and provide an example of matching regex in a comment here.

server/validate-color.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Server-side color validation for highlight colors.
+ */
+
+const PRESET_NAMES = new Set([
+  "yellow", "red", "green", "blue", "purple", "pink", "orange", "teal",
+]);
```
csalvato: This set of colors is defined elsewhere. Shouldn't we export it to reuse?

cass-clearly (commented):
## Review feedback addressed — all 7 points

### 1. Remove teal and pink from palette ✅
Removed teal and pink from COLOR_PRESETS. Palette is now 6 colors: yellow, red, green, blue, purple, orange.

### 2. DEFAULT_COLOR readability (previously unaddressed) ✅
Changed from hardcoded #ffd400 to COLOR_PRESETS["yellow"] — now lives in shared/color-constants.js.

### 3. Rename regex to HEX_COLOR_REGEX with example comment ✅
Renamed HEX_RE to HEX_COLOR_REGEX with comment: // Matches 6-digit hex color codes, e.g. "#ff6b6b"

### 4. CHECK constraint on color column ✅
server/index.js migration now adds a CHECK constraint on the color column allowing NULL, preset names, or valid hex codes. Uses DROP CONSTRAINT IF EXISTS + ADD CONSTRAINT for idempotency.

### 5. Shared hex regex (single location) ✅
Created shared/color-constants.js as the single source of truth. Both feedback-layer/src/utils/color.js (ESM re-export) and server/validate-color.js (CJS require) import from it.

### 6. Shared allowed colors (single location) ✅
PRESET_NAMES is derived from Object.keys(COLOR_PRESETS) in the shared module. Server validate-color.js and index.js error messages both use it — no more hardcoded color lists.

### 7. Update color after creation ✅
PATCH /comments/:id already supports color updates (set to a preset name, hex code, or null to clear). Updated the README table to reflect this and added curl examples.

### README update ✅
- Updated PATCH /comments/:id description: Update body, status, or color
- Added Highlight Colors section with preset list, hex format, and curl examples for create/update/clear

All client tests pass (45/45). Changes in commit 8c0cd42.

csalvato:
There's no way to change the color of an existing comment in the UI. Can we fix?

csalvato:
And this is a comment for testing the pipeline of comments into queue. You can disregard.

cass-clearly (commented):
Addressed the review feedback — users can now change the color of an existing comment directly from the sidebar.

**What changed:**

- **`api.js`**: `updateComment()` now accepts an optional `color` field and forwards it in the PATCH payload (alongside `body`). Only defined fields are sent.
- **`sidebar.js`**: Each root comment card renders a row of small color swatches (reusing the shared `COLOR_PRESETS`). The current color is indicated with a border. Clicking a different swatch fires the new `onColorChange(commentId, hex)` callback. Clicks on swatches are excluded from the card's click-to-scroll handler.
- **`index.js`**: New `handleColorChange()` calls `updateComment` with the new color, removes the old highlight, re-anchors with the new color, and re-renders the sidebar.

Client tests: 45/45 passing. Server tests have a pre-existing DB config issue (SCRAM password not configured in this environment) unrelated to these changes.

csalvato (approved):

PR #131, opened 2026-02-21
Virtual scrolling for 1000+ annotations

### LINKED ISSUE #75: Virtual scrolling for 1000+ annotations
## Jobs to Be Done

- As a **reviewer on a heavily-annotated document**, I want the sidebar to remain responsive even with 1000+ comments, so that I can scroll through feedback without lag.
- As a **document author**, I want to view all feedback on a large document without the browser freezing, so that I can address every comment.

## Technical Proposal

Replace the current "render all comments" approach with virtual scrolling (windowed rendering). Only render the comment cards visible in the sidebar viewport plus a small buffer above and below. Use a fixed-height estimate per card (or measure dynamically) to calculate scroll height and which items to render. Implement using either a lightweight virtual scroll library or a custom implementation using `IntersectionObserver`. As the user scrolls, swap out off-screen cards and render on-screen ones. This keeps the DOM node count constant regardless of total comment count. The comments array stays fully loaded in memory for search/filter — only the rendering is virtualized. Client-side change only.

cass-clearly:
## Summary

- Adds windowed rendering to the annotation sidebar so only visible threads (plus a buffer) are in the DOM at any time, keeping node count constant regardless of total comment count
- Extracts pure `computeVisibleRange` and `scrollOffsetForItem` functions into `feedback-layer/src/utils/virtual-scroller.js` for testability
- Thread heights are estimated on first render (80px per card), then measured and cached for pixel-accurate scroll positioning
- `focusCommentCard` updated to scroll-to-and-render the target thread before highlighting, ensuring off-screen comments can be focused from highlight clicks
- CSS changed from flex gap to per-thread margin so spacer divs don't introduce unwanted gaps

Closes #75

## Test plan

- [x] All 44 client tests pass (14 new tests for `computeVisibleRange` and `scrollOffsetForItem`)
- [x] 100% line/branch/function coverage on `virtual-scroller.js`
- [x] Bundle builds successfully
- [ ] Manual: load a page with 1000+ annotations, verify smooth scrolling and constant DOM node count
- [ ] Manual: click a highlight in the document, verify sidebar scrolls to and highlights the correct card
- [ ] Manual: toggle "Show closed" filter, verify virtual scroll updates correctly
- [ ] Manual: add/edit/delete/reply to comments, verify re-render works


cass-clearly (commented):
## Review: Virtual scrolling for annotation sidebar (Round 2)

**Verdict: Changes requested** (posting as comment because GitHub blocks request-changes on own PRs)

### Test results
- **Virtual scroller tests:** 14/14 pass, 100% coverage on `virtual-scroller.js`
- **All other client tests:** 61/63 pass (2 failures are in `markdown.test.mjs`, discussed below)
- **Server tests:** Cannot run (no PostgreSQL in this environment), not affected by this PR

### Previous review bugs: VERIFIED FIXED

Both bugs from the first review are resolved correctly:

1. **Filter toggle ordering** -- `_lastCommentRanges` is now stored and passed through on re-render. Verified that all external callers in `index.js` pass ranges explicitly, and the internal filter-toggle re-render on line 137 passes `_lastCommentRanges`. The `if (commentRanges.size > 0)` guard on line 224 prevents accidental overwrite.

2. **Height cache invalidation** -- `_heightCache.clear()` on line 225 runs at the top of every `renderComments` call. Simple and correct.

### BLOCKING: Missing `markdown.js` dependency

**This PR will crash at runtime.** The diff adds:
- `import { renderMarkdown } from "./utils/markdown.js"` (line 11)
- `renderMarkdown(ann.body)` in `buildCard` (line 418)
- `renderMarkdown(originalText)` in edit-cancel handler (line 545)
- Markdown format hints UI in comment/reply forms (lines 169, 472)
- CSS for `code`, `a`, and `.fb-fmt-hints` styles

**But `feedback-layer/src/utils/markdown.js` does not exist on the `feat/virtual-scroll-sidebar` branch.** It only exists on the local `feat/markdown-comments` branch. This means the bundle will fail to build and the sidebar will crash with an unresolved import error.

**Fix options:**
- (a) Add `markdown.js` and its tests to this PR branch, or
- (b) Revert the markdown-related changes and keep this PR focused on virtual scrolling only. Replace `renderMarkdown(ann.body)` with `escapeHtml(ann.body)` (the original behavior), remove the `renderMarkdown` import, remove the `.fb-fmt-hints` UI elements, and remove the markdown CSS rules.

Option (b) is cleaner -- ship markdown support as its own PR.

### Virtual scrolling implementation: solid

The core virtual scrolling logic is well done:
- Clean extraction of `computeVisibleRange` and `scrollOffsetForItem` into a testable utility module, following the project's established `utils/` pattern
- Correct `rAF` debouncing in `_onScroll` with the `if (_rafId) return` guard
- Correct range-skip optimization to avoid redundant DOM work
- The estimate-then-measure height strategy is standard and implemented correctly
- `focusCommentCard` correctly handles virtual scrolling: compute offset, set scrollTop, force render by clearing `_lastRange`, then query DOM
- `_measureRenderedThreads` recalculates spacer heights after measurement changes

### Minor observations (non-blocking, for future consideration)

1. **`CSS.escape` for comment IDs in selectors** -- `focusCommentCard` interpolates `commentId` directly into `.fb-cmt-card[data-id="${commentId}"]`. IDs are server-generated with a controlled `cmt_*` format so this is safe today, but `CSS.escape(commentId)` would be more defensive.

2. **No `cancelAnimationFrame` cleanup** -- If the sidebar is torn down, `_rafId` isn't cancelled. Not a practical issue with current usage (sidebar lives for the page lifetime) but a latent resource leak.

3. **`_threadData = []` on empty states** -- Done correctly on lines 252 and 261, preventing stale virtual scroll state from persisting.

cass-clearly:
## 🧪 Test This PR

```bash
./scripts/preview.sh 131
```

This checks out the branch, builds, and starts the dev server on http://localhost:3333. Press Ctrl+C to stop and return to main.

PR #141, opened 2026-02-21
Add admin dashboard for managing annotations (#85)

### LINKED ISSUE #85: Admin dashboard for managing annotations
## Jobs to Be Done

- As a **self-hosting admin**, I want a dashboard to view and manage all documents and annotations, so that I can monitor usage and moderate content.
- As a **document author**, I want a central view of all my documents and their annotation counts, so that I can see which documents need attention.

## Technical Proposal

Build a lightweight admin UI served at a `/admin` route on the server. Use server-rendered HTML (no framework — keep it simple with template literals or a minimal template engine) with the existing Express server. Pages: document list (with comment counts and last activity), document detail (showing all annotations), and a moderation view for flagging/deleting spam. Authentication for the admin panel via a simple password set through an environment variable (`ADMIN_PASSWORD`). The admin UI consumes the same REST API internally. Style it simply with a minimal CSS framework or inline styles matching Remarq's visual identity. No client-side JavaScript framework needed — progressive enhancement with form submissions.

cass-clearly:
## Summary

- Adds a server-rendered admin dashboard at `/admin` for managing documents and moderating comments
- Password-protected via `ADMIN_PASSWORD` env var with session-based auth and CSRF protection
- Document list shows all documents with comment counts and last activity timestamps
- Document detail page shows all annotations with threaded replies, status badges, and blockquoted text
- Comment moderation: delete comments with a reason, logged to `moderation_log` table
- Simple, responsive UI matching Remarq branding (purple theme, clean design)
- No client-side JS framework — progressive enhancement with form submissions
- 77 tests passing, 98% line coverage

## New files

- `server/middleware/auth.js` — Session auth, CSRF protection, timing-safe password checking
- `server/routes/admin.js` — Admin routes (login, documents list, document detail, comment deletion)
- `server/views/admin.js` — Server-rendered HTML templates using template literals
- `server/public/admin.css` — Admin dashboard styling
- `.env.example` — Documents ADMIN_PASSWORD and other env vars

## Modified files

- `server/index.js` — Mounts admin router, adds session middleware, creates moderation_log table
- `server/package.json` — Adds express-session dependency
- `server/test.mjs` — Admin dashboard integration tests (auth, CSRF, moderation logging)
- `README.md` — Admin dashboard documentation

## Test plan

- [x] All 77 tests pass (`npm run test:server`)
- [x] Coverage at 98% (threshold: 80%)
- [ ] Verify `/admin` redirects to login when unauthenticated
- [ ] Verify login with correct/incorrect password
- [ ] Verify document list shows documents with comment counts
- [ ] Verify document detail shows comments with delete forms
- [ ] Verify comment deletion logs to moderation_log
- [ ] Verify CSRF protection blocks requests without tokens
- [ ] Verify logout destroys session

Closes #85

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -15,6 +15,7 @@ app.use(cors());
 app.use(express.json());
 
 app.use(
+  "/admin",
   session({
     secret: process.env.SESSION_SECRET || randomBytes(32).toString("hex"),
     resave: false,
--- a/server/middleware/auth.js
+++ b/server/middleware/auth.js
@@ -17,11 +17,12 @@ function requireAuth(req, res, next) {
 function verifyCsrf(req, res, next) {
   const token = req.body && req.body._csrf;
   const sessionToken = req.session && req.session.csrfToken;
-  if (
-    !token ||
-    !sessionToken ||
-    !timingSafeEqual(Buffer.from(token), Buffer.from(sessionToken))
-  ) {
+  if (!token || !sessionToken) {
+    return res.status(403).send("Invalid CSRF token");
+  }
+  const a = Buffer.from(token);
+  const b = Buffer.from(sessionToken);
+  if (a.length !== b.length || !timingSafeEqual(a, b)) {
     return res.status(403).send("Invalid CSRF token");
   }
   next();
--- a/server/routes/admin.js
+++ b/server/routes/admin.js
@@ -11,12 +11,11 @@ function createAdminRouter(pool) {
 
   // ── Login ──────────────────────────────────────────────────────────
 
-  router.get("/login", (req, res) => {
+  router.get("/login", (_req, res) => {
     if (!process.env.ADMIN_PASSWORD) {
       return res.status(503).send("ADMIN_PASSWORD not configured");
     }
-    const csrf = ensureCsrfToken(req);
-    res.send(loginPage(null, csrf));
+    res.send(loginPage());
   });
 
   router.post("/login", (req, res) => {
@@ -24,8 +23,7 @@ function createAdminRouter(pool) {
       req.session.authenticated = true;
       return res.redirect("/admin/documents");
     }
-    const csrf = ensureCsrfToken(req);
-    res.status(401).send(loginPage("Invalid password", csrf));
+    res.status(401).send(loginPage("Invalid password"));
   });
 
   router.post("/logout", (req, res) => {
--- a/server/views/admin.js
+++ b/server/views/admin.js
@@ -42,7 +42,7 @@ function layout(title, content) {
 </html>`;
 }
 
-function loginPage(error, csrfToken) {
+function loginPage(error) {
   return `<!DOCTYPE html>
 <html lang="en">
 <head>
@@ -56,7 +56,6 @@ function loginPage(error, csrfToken) {
     <h1>Remarq Admin</h1>
     ${error ? `<p class="error">${escapeHtml(error)}</p>` : ""}
     <form method="POST" action="/admin/login">
-      <input type="hidden" name="_csrf" value="${csrfToken || ""}">
       <label for="password">Password</label>
       <input type="password" id="password" name="password" required autofocus>
       <button type="submit">Login</button>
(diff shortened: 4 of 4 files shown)


cass-clearly (commented):
## Re-review — PR #141: Add admin dashboard for managing annotations

**Verdict: Approve** (posting as comment because GitHub won't let me approve the repo owner's PR)

### Test Results

All tests pass:
- **77 server tests**: PASS (98.1% line coverage)
- **30 client tests**: PASS (100% line coverage)

Coverage well above the 80% threshold.

### Verification of Previous Review Fixes

All four issues from the prior review have been addressed correctly in commit `2183bc2`:

**1. CSRF length-mismatch bug (was: blocking)** — Fixed. `verifyCsrf` now checks `a.length !== b.length` before calling `timingSafeEqual`, returning a clean 403 instead of throwing a 500. This matches the pattern already used in `checkPassword`. Correct fix.

**2. Misleading CSRF field on login form (was: blocking)** — Fixed. The hidden `_csrf` field has been removed from the login form. The `loginPage` function no longer accepts a `csrfToken` parameter, and the login routes no longer call `ensureCsrfToken`. No dead code left behind. Clean resolution — removing the theater rather than adding enforcement to the login POST is the right call since there's no session to protect pre-authentication.

**3. Unescaped `comment.status` in CSS class (was: non-blocking)** — Fixed. The class interpolation now reads `status-${escapeHtml(comment.status)}` and the display text is also escaped. Defense-in-depth restored.

**4. Session middleware scoped to /admin (was: non-blocking)** — Fixed. The session middleware in `server/index.js` is now mounted on `"/admin"` rather than globally. JSON API endpoints no longer incur session overhead.

### Remaining Observations (Non-blocking, no action required)

- The CSRF token in the hidden field (`value="${csrfToken}"`) is not passed through `escapeHtml`. This is safe because `randomBytes(24).toString("base64url")` can only produce `[A-Za-z0-9_-]` characters, but wrapping it in `escapeHtml` would be a zero-cost defense-in-depth measure for future-proofing if the token generation ever changes.
- The duplicated `asyncHandler` in `routes/admin.js` vs `index.js` is still present — still fine for two files, still worth extracting if a third consumer appears.
- No rate limiting on login — acknowledged in prior review as acceptable for v1.

### Summary

The security fixes are correct and complete. The code is clean, well-tested, and appropriately scoped. Ship it.

cass-clearly (commented):
## Review: Approve (posted as comment — same-user restriction on GH approval)

**Tests:** All 77 tests pass, 98.1% line coverage (threshold: 80%). Solid integration test coverage of auth flow, CSRF validation, moderation logging, and session lifecycle.

### Security

- **Password check:** `timingSafeEqual` used correctly. Early exit on length mismatch leaks password length, but acceptable for a single shared admin password — hashing would be overengineering here.
- **CSRF:** Properly applied to the only destructive authenticated action (comment deletion). Token comparison uses `timingSafeEqual` with length check. Login form intentionally omits CSRF — password itself is proof of intent, and login CSRF is a non-concern for a single-password admin panel.
- **Sessions:** Scoped to `/admin` routes only (no overhead on API requests). `httpOnly`, `sameSite: "lax"`, `saveUninitialized: false` — correct defaults. Auto-generated secret when `SESSION_SECRET` is unset.
- **XSS:** All user-controlled data passes through `escapeHtml()`. Covers `&`, `<`, `>`, `"`. No single-quoted attribute delimiters in templates, so missing `'` escaping is safe today. Integer-typed `comment_count` rendered unescaped is also safe (comes from `COUNT()::int`).
- **SQL injection:** All queries use parameterized statements. No string interpolation in SQL. Clean.

### Architecture

- **Server-rendered HTML with template literals:** Right choice. No framework dependency for an admin panel that's mostly read-only views with a few forms. Progressive enhancement via `onsubmit="return confirm(...)"` is appropriate.
- **Session scoping:** Session middleware mounted only on `/admin` path — doesn't burden the public API. Good.
- **Moderation log without FKs:** Correct — audit logs should survive referenced entity deletion.
- **`asyncHandler` duplicated** in `index.js` and `routes/admin.js`: One-liner, not worth extracting to shared module for two call sites. Fine.
- **In-memory session store (MemoryStore default):** Appropriate for an admin panel with few concurrent users. Not worth a session store dependency at this stage.

### Notes (non-blocking)

1. **No rate limiting on `POST /admin/login`:** Unlimited brute-force attempts are possible. Consider adding basic rate limiting (e.g., `express-rate-limit`) in a follow-up if this will be exposed to the internet. Not blocking because password-in-env-var implies trusted-network deployment.
2. **`escapeHtml` doesn't escape `'`:** Safe today since all attribute values use double quotes, but worth adding `'` → `&#39;` for defense-in-depth if the template grows.
3. **No transaction on comment deletion:** The two `DELETE` queries (replies then parent) in `routes/admin.js:89-90` are not wrapped in a transaction, but this is consistent with the existing `DELETE /comments/:id` endpoint and the failure mode is benign (moderation log entry exists, comment deletion partially fails — no orphan state).

### PR Description Accuracy

Description is accurate. One minor note: it says "CSRF protection on all actions" but CSRF is only on the delete endpoint (which is the only state-mutating authenticated action besides login/logout). This is the correct behavior — just a slight overstatement in the description.

Clean, pragmatic PR. No overengineering, no significant underengineering. **Approved.**

Issue #1, opened 2026-02-17
Rename project to Remarq


cass-clearly:
## Rename: Glossator → Remarq

**Why:** "Glossator" is too esoteric — it's a medieval term most people won't recognize. The tool has outgrown its original scope; it's not just an annotation layer, it's a collaborative document platform where AI is a first-class participant.

**Why Remarq:**
- **Remark** — leaving comments/feedback
- **Mark** — highlighting, annotating text  
- **Re-mark** — revising, iterating on a document
- The **q** gives it a distinctive, trademarkable look
- Scales from annotation tool to full document platform

**What needs to change:**
- [ ] Rename GitHub repo to `remarq`
- [ ] Update package name in `package.json` files
- [ ] Update all references in README, code comments, and docs
- [ ] Check domain availability (`remarq.dev`, `remarq.io` look promising; `remarq.ai` and `remarq.com` are taken)
- [ ] Update npm package name if published

Issue #60, opened 2026-02-20
Private vs. public annotations


cass-clearly:
## Jobs to Be Done

- As a **reviewer**, I want to leave private annotations only I can see, so that I can take personal notes on a document without sharing them with the team.
- As a **document author**, I want to control whether annotations on my document are public or private by default, so that I can manage the visibility of feedback.

## Technical Proposal

Add a `visibility` column to the comments table with values `public` (default) and `private`. Private comments are filtered out of API responses unless the requester identifies as the comment author. On the client side, add a toggle (lock icon) in the comment creation form to mark a comment as private. Private annotations render with a distinct highlight style (e.g., dashed border). The `GET /comments` endpoint gains a `visibility` query param and checks authorship for private comments. This requires some form of user identity — either the existing name-based system (weak but functional) or integration with the SSO feature (#41) for stronger guarantees.

PR #149, opened 2026-02-22
Add private vs. public annotations (#60)

cass-clearly:
## Summary
- Adds `visibility` column (public/private) to the comments table with a safe idempotent migration
- **GET /comments**: New `?viewer=<name>` param filters out private comments unless the viewer is the author
- **POST /comments**: Accepts optional `visibility` field, validates values
- **PATCH /comments/:id**: Allows toggling visibility between public and private
- Client: comment form has a lock-icon toggle for private mode, private comments show dashed borders on highlights and dashed card borders with a lock badge in the sidebar
- 9 new server integration tests covering create, filter, and update visibility

## Test plan
- [x] Server tests pass: `DATABASE_URL=... npm run test:server` — 137 tests, 98% coverage
- [x] Client tests pass: `npm run test:client` — 41 tests, 100% coverage
- [ ] Manual: create a private annotation, verify it has dashed highlight and lock badge
- [ ] Manual: verify another user (different name) cannot see the private annotation
- [ ] Manual: toggle visibility from private to public via PATCH

### DIFF
--- a/feedback-layer/src/sidebar.js
+++ b/feedback-layer/src/sidebar.js
@@ -217,7 +217,7 @@ export function showCommentForm(quote) {
         <button class="fb-btn fb-btn-primary fb-submit-btn">Add Comment</button>
         <button class="fb-btn fb-btn-cancel fb-cancel-btn">Cancel</button>
         <button class="fb-visibility-toggle" title="Toggle private/public" data-visibility="public">
-          <svg class="fb-visibility-icon-public" xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><rect x="3" y="11" width="18" height="11" rx="2" ry="2"/><path d="M7 11V7a5 5 0 0 1 10 0v4"/></svg>
+          <svg class="fb-visibility-icon-public" xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><rect x="3" y="11" width="18" height="11" rx="2" ry="2"/><path d="M7 11V7a5 5 0 0 1 9.9-1"/></svg>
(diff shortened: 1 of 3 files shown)


cass-clearly (commented):
## Review: Add private vs. public annotations (#60)

**Tests:** All 178 tests pass (137 server + 41 client). Coverage is 98%/100%. No regressions.

PR description is accurate. The implementation is clean overall — solid validation, idempotent migration, good test coverage. But there are a few real issues that need addressing before merge.

---

### Must fix

**1. `GET /comments/:id` bypasses visibility filtering (bug)**

The list endpoint (`GET /comments`) correctly filters by visibility, but the single-resource endpoint (`GET /comments/:id`, server/index.js L437-453) returns any comment by ID regardless of visibility. A private comment's ID can be fetched directly by anyone:

```
curl http://localhost:3333/comments/cmt_abc123
```

This is inconsistent with the list endpoint. Add the same visibility check here, or at minimum require a `viewer` param.

**2. Both visibility toggle SVG icons are identical (bug)**

Lines 219-221 of `sidebar.js`: the "public" and "private" SVGs use the exact same `<path>` and `<rect>` elements (both are unlocked padlocks). The toggle changes the background color via CSS, but both icons are visually identical. The private icon should be visually distinct (e.g., a closed/locked padlock SVG where the shackle is not open).

---

### Should fix

**3. Document the security model limitation**

The `viewer` param is unauthenticated — any HTTP client can pass `?viewer=Alice` to see Alice's private comments. This is consistent with the existing name-based identity model (no auth for non-tenant users), but the word "private" sets user expectations that the server doesn't meet. At minimum, add a code comment near the visibility filter noting that this is a UI-level visibility feature, not an access-control boundary, so future contributors don't assume otherwise.

---

- `updateComment` in `api.js` now takes a generic `fields` object but only forwards `body` and `visibility`. If someone passes `status` or `sort_order` through it, they'd be silently dropped. Not a bug today since separate functions exist for those, but the generic parameter name is a maintenance hazard.

- PATCH runs up to 4 separate UPDATE queries without a transaction. Pre-existing issue, but this PR adds a 4th. Not blocking.

- Replies to a private comment are not filtered — they'd appear as orphans if the parent is hidden. Edge case worth considering in a follow-up.

---

**Verdict: Request changes.** Fix issues #1 and #2 before merge. The rest are non-blocking.

PR #151, opened 2026-02-22
Add interactive demo site (#87)

### LINKED ISSUE #87: Interactive demo site
Migrated from TODO.md.

cass-clearly:
## Summary
- Adds a demo banner to `serve/index.html` that activates on `demo.remarq.dev` or with `?demo` query parameter, explaining what Remarq is and linking to the GitHub repo
- Implements session-based sandbox isolation: each browser session generates a unique document URI via `sessionStorage`, so demo visitors don't see each other's annotations
- Adds `server/demo-cleanup.js` with TTL-based periodic cleanup (deletes comment threads older than 24 hours, plus orphaned documents) — activated when `DEMO_MODE=true` env var is set
- Integrates cleanup into server startup via `startDemoCleanup(pool)` in `server/index.js`
- Adds 6 tests covering: TTL expiry, reply cascade, orphaned document cleanup, mixed age preservation, empty state, and reaction cascade

## How it works
**Client-side (demo page):**
- Demo mode detected via `location.hostname === 'demo.remarq.dev'` or `?demo` query param
- When active: shows a fixed purple banner, adds body padding, and generates a `sessionStorage`-based random document URI for sandbox isolation
- Non-demo mode: banner hidden, feedback-layer loads normally (no behavioral change)

**Server-side (TTL cleanup):**
- `DEMO_MODE=true` env var triggers `startDemoCleanup(pool)` on server start
- Runs cleanup immediately on boot, then every hour via `setInterval` (unref'd so it doesn't keep process alive)
- Deletes replies of expired parents first (FK constraint), then expired top-level comments, then orphaned documents
- Reactions cascade automatically via `ON DELETE CASCADE`

## Test plan
- [x] All 187 tests pass (146 server + 41 client)
- [x] Server coverage above 80% threshold
- [ ] Verify demo banner shows on `?demo` query param locally
- [ ] Verify sandbox isolation creates unique document URIs per session
- [ ] Verify non-demo mode is unaffected (no banner, same-origin document URI)

Closes #87

### DIFF
--- a/serve/index.html
+++ b/serve/index.html
@@ -37,6 +37,28 @@
       padding: 20px;
       margin: 2em 0;
     }
+    #remarq-demo-banner {
+      display: none;
+      background: linear-gradient(135deg, #7c3aed, #6d28d9);
+      color: #fff;
+      padding: 12px 20px;
+      text-align: center;
+      font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, sans-serif;
+      font-size: 0.95em;
+      line-height: 1.5;
+      position: fixed;
+      top: 0;
+      left: 0;
+      right: 0;
+      z-index: 10000;
+    }
+    #remarq-demo-banner a {
+      color: #e0d5f5;
+      text-decoration: underline;
+      font-weight: 600;
+    }
+    #remarq-demo-banner a:hover { color: #fff; }
+    body.remarq-demo-active { padding-top: 50px; }
   </style>
   <script src="https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.min.js"></script>
   <script>
@@ -55,6 +77,11 @@
   </script>
 </head>
 <body>
+
+<div id="remarq-demo-banner">
+  <strong>Remarq Demo</strong> — Select any text below to try the annotation layer.
+  <a href="https://github.com/cass-clearly/remarq">View on GitHub</a>
+</div>
 
 <article>
   <h1>The Future of Collaborative Document Editing</h1>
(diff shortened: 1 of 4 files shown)


cass-clearly (commented):
## Review: Request Changes

**Tests pass.** All 187 tests pass (146 server + 41 client), coverage above the 80% threshold. The code is clean and well-structured overall. However, there are two data-safety bugs that need to be fixed before merge.

---

### BUG (Critical): Cleanup queries are globally scoped — will destroy non-demo data

Both `cleanupDemoComments` queries operate on the entire `comments` and `documents` tables with no scoping. If `DEMO_MODE=true` is set on a server that also holds self-hosted or multi-tenant data, these queries will:

1. **Delete any old comments across ALL tenants**, not just demo-created ones:
   ```sql
   DELETE FROM comments WHERE parent IS NULL AND created_at < $1
   ```
   This nukes every comment older than 24h, regardless of origin.

2. **Delete any document with zero comments across ALL tenants**:
   ```sql
   DELETE FROM documents WHERE id NOT IN (
     SELECT DISTINCT document FROM comments
   )
   ```
   Any tenant document that happens to have no comments gets deleted.

**Fix:** Scope the cleanup to demo data only. The client-side already generates document URIs with the `https://demo.remarq.dev/` prefix. Use that to scope:

```sql
-- Only delete comments belonging to demo documents
DELETE FROM comments WHERE parent IN (
  SELECT c.id FROM comments c
  JOIN documents d ON c.document = d.id
  WHERE c.parent IS NULL AND c.created_at < $1 AND d.uri LIKE 'https://demo.remarq.dev/%'
)
```

And similarly for the top-level comment and orphaned document cleanup queries. Alternatively, add an `api_key` or `is_demo` flag to documents. The URI prefix approach is the simplest since the client already enforces it.

---

### Issue (Minor): Test suite coupling via shared pool

The `pool.end()` call was removed from the multi-tenant suite's `after()` hook and relocated to the demo-cleanup suite's `after()`. This creates an ordering dependency between test suites — if the demo-cleanup suite is ever skipped or moved, the pool connection leaks. Consider using a single global teardown hook (e.g., a top-level `after()`) that calls `pool.end()`.

---

### Observation (Non-blocking): Weak sandbox isolation entropy

`Math.random().toString(36).slice(2, 10)` produces ~41 bits of entropy. For a public demo this is fine — the "isolation" is a UX convenience, not access control, and the PR doesn't claim otherwise. Just noting it for awareness: a determined user could enumerate active demo sessions and read their annotations. If this ever matters, switch to `crypto.randomUUID()`.

---

### What's good

- The client-side demo detection (`hostname` check + `?demo` query param) is clean and doesn't affect non-demo behavior
- The `timer.unref?.()` pattern is correct — won't keep the process alive
- Error handling in `startDemoCleanup` is solid (catch on both the immediate run and the interval)
- Tests cover the important cases: TTL expiry, reply cascade, orphaned cleanup, mixed-age preservation, empty state, and reaction cascade
- The banner UX (fixed position, body padding, z-index) is well thought out

**Bottom line:** The globally-scoped cleanup queries are a data-safety hazard. Scope them to demo documents (by URI prefix or a dedicated flag) and this is good to merge.

PR #170, opened 2026-02-23
chore: pre-commit hook to auto-rebuild feedback-layer bundle

cass-clearly:
## Summary

- Adds `.githooks/pre-commit` hook that detects staged changes in `feedback-layer/src/`, runs the build, and auto-stages the updated `serve/feedback-layer.js` bundle
- Adds `prepare` script to root `package.json` so `npm install` automatically configures `core.hooksPath`
- Documents the build step requirement and hook setup in `CLAUDE.md`

## Why

PR #153 shipped with a stale bundle because the build step was skipped. PR #169 added documentation as a safeguard, but a pre-commit hook is strictly better — it enforces the rebuild automatically rather than relying on developers/agents to remember.

## How to verify

1. `npm install` (should configure hooks path automatically)
2. Edit any file in `feedback-layer/src/`, stage it, and commit — the hook should rebuild and stage the bundle
3. Commit a file outside `feedback-layer/src/` — the hook should exit silently

Supersedes #169.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -48,10 +48,6 @@ If the review agent requests changes, fix them on the same branch, push, and re-
 ```
 gh pr merge --squash --delete-branch
 ```
-  
-## Documentation
-
-All functional changes (new features, changed behavior, new keyboard shortcuts, configuration options) must be documented in README.md before the PR is opened.
 
 ## Testing & Coverage
 


CLAUDE.md
```diff
@@ -45,6 +45,22 @@ If the review agent requests changes, fix them on the same branch, push, and re-
...
+```
+
+This updates `serve/feedback-layer.js` which is what the server actually serves. If you skip this step, your changes will exist in source but won't be visible to users. Always commit the rebuilt `serve/feedback-layer.js` alongside your source changes.
+
+A pre-commit hook in `.githooks/pre-commit` handles this automatically — it detects staged changes in `feedback-layer/src/`, runs the build, and stages the updated bundle. The hook is enabled by the `prepare` script on `npm install`, or manually:
+
+```bash
+git config core.hooksPath .githooks
+```
+
```
csalvato: Unnecessary context now that we have a hook.

package.json
```diff
@@ -5,6 +5,7 @@
   "license": "AGPL-3.0",
   "description": "Drop-in annotation and comment layer for any HTML page.",
   "scripts": {
+    "prepare": "git config core.hooksPath .githooks",
```
csalvato: what's this do?

CLAUDE.md
cass-clearly: Done — replaced the verbose build docs with a 3-line note that just says the hook handles it and `npm install` enables it via `prepare`.

package.json
cass-clearly: The `prepare` script is a built-in npm lifecycle hook — it runs automatically after `npm install`. It sets `core.hooksPath` so git uses our `.githooks/` directory for hooks (like the pre-commit build hook). This way anyone who clones and runs `npm install` gets the hooks enabled automatically without manual setup.

csalvato (approved):

csalvato (approved):

PR #193, opened 2026-02-24
fix: inject print-hide CSS to fix iOS Safari print preview

### LINKED ISSUE #191: Sidebar and highlights visible in iOS Safari print preview
## Problem
The Remarq sidebar (with all comments) and text highlights appear in print preview on iOS Safari (iPhone). Desktop browsers correctly hide them via `@media print` CSS in the host page.

## Expected
All Remarq UI elements (sidebar, tab, highlights, tooltips) should be hidden when printing, on all browsers including mobile Safari.

## Suggested Fix
Inject print-hide CSS directly in the feedback-layer JS bundle during initialization, rather than relying on host pages to add it manually. Something like:

```js
const printStyle = document.createElement('style');
printStyle.textContent = `
  @media print {
    .fb-sidebar, .fb-sidebar-tab, .fb-annotate-tooltip,
    [class^="fb-"], [class*=" fb-"],
    [class^="hf-"], [class*=" hf-"] {
      display: none !important;
      visibility: hidden !important;
    }
  }
`;
document.head.appendChild(printStyle);
```

This ensures it works everywhere without requiring each host page to add print CSS.

## Reproduction
1. Open any page with Remarq sidebar on iPhone Safari
2. Tap Share → Print
3. Sidebar and comments appear on page 1 of print preview

## Environment
- iOS Safari (iPhone)
- Remarq v2.0.0

cass-clearly:
## What changed

Injects `@media print` CSS directly during feedback-layer initialization to hide all Remarq UI elements (sidebar, tab, highlights, tooltips) when printing.

- Added `feedback-layer/src/utils/print-hide.js` utility that injects a `<style>` element targeting all `fb-` and `hf-` prefixed classes
- Called `injectPrintHideStyles()` at the start of `boot()` in `feedback-layer/src/index.js`
- Added test coverage for the print-hide injection
- Rebuilt bundle

## Why

Closes #191

iOS Safari ignores `@media print` CSS rules in the host page's stylesheet, causing Remarq UI elements (sidebar, tab, highlights, tooltips) to appear in print preview. By injecting the print-hide styles directly from the JS bundle during initialization, it works across all browsers without requiring host pages to add print CSS.

## How to verify

1. Run client tests: `npm run test:client` — all 46 tests pass
2. Load a page with the feedback-layer in iOS Safari
3. Open print preview — sidebar, tab, highlights, and tooltips should be hidden
4. Verify desktop browsers still hide UI elements in print preview

## Manual testing checklist

- [x] Existing tests pass (`npm run test:client`)
- [ ] Tested in browser (annotations, sidebar, highlights work)
- [ ] No console errors in browser DevTools


cass-clearly (commented):
## Review: Approved (posted as comment — cannot self-approve on GitHub)

Clean, minimal fix. The implementation is correct and well-scoped.

### What I verified

- **All 46 client tests pass**, including the new `injectPrintHideStyles` test. 100% line coverage on print-hide.js.
- **Bundle is rebuilt** — `serve/feedback-layer.js` contains the `data-remarq-print` style injection and `@media print` rule.
- **CSS selectors are correct** — `[class^="fb-"]`, `[class*=" fb-"]`, `[class^="hf-"]`, `[class*=" hf-"]` covers all Remarq UI elements (sidebar, tab, highlights, tooltips from `fb-` prefix; AI revision modal from `hf-` prefix).
- **Timing is safe** — `injectPrintHideStyles()` runs inside `boot()`, which fires after DOMContentLoaded, so `document.head` is guaranteed to exist.
- **PR description accurately reflects the changes** and follows the template. No API endpoints changed (section correctly omitted).

### One minor nit (non-blocking)

The test shims `globalThis.document` but restores it *after* assertions rather than in a `finally` block or `afterEach`. If an assertion were to throw, the mock document leaks into subsequent tests. In practice this is harmless here (only `api` tests follow, and they don't touch `document`), but a `try/finally` would be more defensive:

```js
it("creates a style element with print media rule", () => {
  const appended = [];
  globalThis.document = { /* shim */ };
  try {
    const style = injectPrintHideStyles();
    // assertions...
  } finally {
    globalThis.document = origDocument;
  }
});
```

Not worth blocking on — ship it.

PR #198, opened 2026-02-26
feat: add per-IP rate limiting to API endpoints

### LINKED ISSUE #186: Add rate limiting to API endpoints
## Problem

No rate limiting exists on any API endpoint. This leaves the server open to:

- Comment/reaction spam
- Denial-of-service via request flooding
- Brute-force document URI enumeration

## Recommendation

Install `express-rate-limit` and apply sensible per-IP limits:

```bash
npm install express-rate-limit
```

```js
const rateLimit = require('express-rate-limit');

app.use(rateLimit({
  windowMs: 15 * 60 * 1000, // 15 minutes
  max: 100,
  standardHeaders: true,
  legacyHeaders: false,
}));
```

Consider tighter limits on write endpoints (POST/PUT/DELETE).

cass-clearly:
## What changed

Add per-IP rate limiting to all API endpoints using `express-rate-limit`. Reads (GET) are limited to 300 requests/min and writes (POST/PATCH/DELETE) are limited to 30 requests/min.

New middleware module `server/rate-limit.js` creates two separate rate limiters dispatched by HTTP method. Responses include standard `RateLimit-*` headers and return `429 Too Many Requests` with the existing error format when exceeded.

## Why

Closes #186. Protects the API against abuse and accidental flooding without impacting normal usage patterns.

## New/Changed Endpoints

No new endpoints. All existing endpoints now include rate limiting headers:

| Header | Description |
|--------|-------------|
| `RateLimit-Limit` | Max requests allowed in the window |
| `RateLimit-Remaining` | Requests remaining in the current window |
| `RateLimit-Reset` | Seconds until the window resets |

A `429` response is returned when the limit is exceeded:
```json
{ "error": { "message": "Too many requests, please try again later" } }
```

## How to verify

1. Run the test suite:
   ```bash
   npm run test:server
   ```
   Five new rate limiting tests verify headers, limits, 429 behavior, and independent tracking.

2. Manual curl test:
   ```bash
   # Check rate limit headers on a GET
   curl -i http://localhost:3333/health
   # Should see RateLimit-Limit: 300

   # Check rate limit headers on a POST
   curl -X POST http://localhost:3333/documents \
     -H "Content-Type: application/json" \
     -d '{"uri":"https://example.com/test"}' -i
   # Should see RateLimit-Limit: 30
   ```

## Manual testing checklist

- [ ] Server starts without errors (`npm run start`)
- [ ] Existing tests pass (`npm test`)
- [ ] Tested API changes with curl (include example commands above)

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,17 @@ Set a default highlight color for all new comments on a page using the `data-def
 
 Accepts any preset name or hex code. Users can still override the color per-comment using the color picker in the sidebar.
 
+### Rate Limiting
+
+All API endpoints are rate-limited per IP address:
+
+| Request type | Limit |
+|-------------|-------|
+| Read (`GET`) | 300 requests/min |
+| Write (`POST`/`PATCH`/`DELETE`) | 30 requests/min |
+
+Rate-limited responses return `429 Too Many Requests` with the standard error format. Responses include `RateLimit-Limit`, `RateLimit-Remaining`, and `RateLimit-Reset` headers.
+
 ## Features
 
 - **No accounts** — reviewers just type their name
--- a/server/index.js
+++ b/server/index.js
@@ -11,6 +11,7 @@ const path = require("path");
 const { rateLimitMiddleware } = require("./rate-limit.js");
 
 const app = express();
+app.set("trust proxy", 1);
 app.use(cors());
 app.use(express.json());
 app.use(rateLimitMiddleware);
--- a/server/index.js
+++ b/server/index.js
@@ -8,10 +8,12 @@ const { sanitize } = require("./sanitize.js");
 const { validateColor } = require("./validate-color.js");
 const { PRESET_NAMES } = require("../shared/color-constants.js");
 const path = require("path");
+const { rateLimitMiddleware } = require("./rate-limit.js");
 
 const app = express();
 app.use(cors());
 app.use(express.json());
+app.use(rateLimitMiddleware);
 
 const DATABASE_URL = process.env.DATABASE_URL || "postgresql://postgres@localhost/postgres";
 const pool = new Pool({ connectionString: DATABASE_URL });
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "cors": "^2.8.5",
         "express": "^4.21.0",
+        "express-rate-limit": "^8.2.1",
         "pg": "^8.13.0"
       },
       "bin": {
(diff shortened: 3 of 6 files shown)


cass-clearly (commented):
## Code Review: Rate Limiting PR

**Tests:** All 5 new rate limiting tests pass. The 1 failing test (`returns 400 for disallowed emoji`) is a pre-existing failure on `main`, not introduced by this PR. Coverage is 100% on rate-limit.js.

### Issues (request changes — posting as comment since GitHub blocks request-changes on own PR)

**1. HEAD and OPTIONS requests hit the write limiter (bug)**

`rateLimitMiddleware` only checks for `req.method === "GET"` — everything else falls through to `writeLimiter`. This means:

- **`HEAD` requests** (semantically identical to GET) count against the 30/min write limit instead of the 300/min read limit. Some monitoring tools and link checkers use HEAD.
- **`OPTIONS` requests** (CORS preflight) count against the write limit. Since `cors()` middleware is applied *before* `rateLimitMiddleware`, Express still processes the preflight through the rate limiter. A browser making cross-origin POST requests will send a preflight OPTIONS first — if a user is near their write limit, the preflight alone could trigger a 429, breaking CORS for legitimate requests.

Suggested fix:

```js
function rateLimitMiddleware(req, res, next) {
  if (req.method === "GET" || req.method === "HEAD" || req.method === "OPTIONS") {
    return readLimiter(req, res, next);
  }
  writeLimiter(req, res, next);
}
```

Or skip rate limiting for OPTIONS entirely if you prefer.

**2. Missing `trust proxy` — rate limiting will break behind a reverse proxy (underengineering)**

The docker-compose file exposes the server on `0.0.0.0:3333`, meaning it will often sit behind a reverse proxy (nginx, Caddy, cloud load balancer) in production. Without `app.set("trust proxy", ...)`, `express-rate-limit` will see every request as coming from the proxy's IP, effectively applying a single shared limit to all users. This can either:
- Block all traffic once any user exceeds the limit, or
- Make the rate limiter useless if the limit is high enough.

`express-rate-limit` v8 [explicitly warns about this](https://express-rate-limit.mintlify.app/guides/troubleshooting#misidentification-of-users-behind-a-proxy). The fix is straightforward:

```js
// In server/index.js, before applying middleware:
app.set("trust proxy", 1); // trust first proxy hop
```

This isn't strictly a blocker for local dev/demo use, but any Docker or cloud deployment will be broken without it. At minimum, document the requirement or wire it to an env var.

### Non-blocking observations

- The module structure is clean and well-separated. No overengineering here.
- Exporting `readLimiter` and `writeLimiter` for test `resetKey()` calls is a good pragmatic choice.
- The 429 error format correctly matches the existing `errorResponse` shape.
- README docs are accurate and well-placed.

### Verdict

Requesting changes for (1) the HEAD/OPTIONS routing bug and (2) the missing `trust proxy` configuration. Both are small fixes.

cass-clearly:
✅ Both review issues resolved:
1. HEAD/OPTIONS now routed to read limiter via READ_METHODS set
2. `trust proxy` configured before middleware

Code is clean. Ready to merge.

cass-clearly (commented):
## Review: feat: add per-IP rate limiting to API endpoints

### What's Good
- Clean separation: `rate-limit.js` module keeps the middleware logic out of `index.js`
- Read/write split (300 vs 30 req/min) is the right approach — matches the issue proposal
- Standard `RateLimit-*` headers included, so clients can backoff gracefully
- Tests updated to cover rate limiting behavior
- README documentation added

1. **`trust proxy` setting**: `app.set('trust proxy', 1)` is required for correct IP detection behind a reverse proxy (nginx, Cloudflare). This is correct, but worth noting in docs that users running behind a proxy should be aware of this.
2. **Rate limit storage**: In-memory store means limits reset on server restart and don't work across multiple instances, but that's clearly acceptable for v1 and the issue explicitly mentioned this.
3. **Note on index.js**: This PR still uses `app.use(cors())` (open CORS) — it was authored before PR #220 merged. These two PRs will have a minor merge conflict on that line. Whichever merges second will need to reconcile.

Solid implementation. Approve.

PR #248, opened 2026-03-08
fix: make npx @csalvato/remarq-server work (REM-16)

cass-clearly:
## What changed

- Moved `color-constants.js` and `emoji-constants.js` from `shared/` to `server/` (canonical location). Left 1-line CJS re-export shims in `shared/` so feedback-layer builds are unchanged.
- Moved `serve/` to `server/serve/` so static assets are included in the npm package.
- Updated `server/package.json` `files` field to include all required modules: `validate-color.js`, `openapi.js`, `color-constants.js`, `emoji-constants.js`, and `serve/`.
- Updated `server/Dockerfile` to match the new layout (no more `COPY shared/` or `COPY serve/`).
- Updated `server/index.js` static path from `path.join(__dirname, "..", "serve")` to `path.join(__dirname, "serve")`.
- Fixed README npx quick-start example to include required `DATABASE_URL` prefix.
- Added automated test that walks the `require()` graph from `index.js` and asserts every local dependency is declared in the `files` field.

## Why

`npx @csalvato/remarq-server` crashed on startup because the npm tarball was missing several required files (`validate-color.js`, `openapi.js`, shared constants, and the `serve/` static assets directory). The `shared/` directory and top-level `serve/` directory couldn't be included in the npm package because npm doesn't follow parent-directory references.

This is part of the README audit (REM-16) — ensuring all quick-start paths are copy-pasteable and actually work.

## How to verify

1. **Run tests:** `npm run test:server` — all 139 pass, including 2 new package completeness tests
2. **Verify npm pack:** `cd server && npm pack --dry-run` — confirm all required files appear in tarball
3. **Verify re-export shims:** `node -e "console.log(require('./shared/color-constants.js'))"` — should print color presets
4. **Verify client build:** `npm run build` — feedback-layer still builds (imports through shims)

## Manual testing checklist

- [x] `npm pack --dry-run` in `server/` includes all required files
- [x] Client tests pass (shared/ re-export shims work)
- [x] MCP and agent-loop tests pass

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -20,7 +20,7 @@ const sharedRules = {
 
 module.exports = [
   // Ignore build output, dependencies, and vendored files
-  { ignores: ["**/node_modules/**", "**/dist/**", "serve/**", "feedback-layer/vendor/**", ".claude/**"] },
+  { ignores: ["**/node_modules/**", "**/dist/**", "server/serve/**", "feedback-layer/vendor/**", ".claude/**"] },
 
   // Base recommended rules for all files
   js.configs.recommended,


cass-clearly:
**[The Craftsperson]** APPROVE

**Test results:** 139 server tests pass, 91 client tests pass. Coverage above 80% threshold on both sides.

**Review through Principles 3, 5, 6, 12:**

**Principle 5 (Tests as thinking tool):** The `collectLocalRequires` test is well-designed — it walks the actual require graph rather than hardcoding a list of expected files. This means the test catches future regressions automatically when someone adds a new `require()` to the server without updating `files`. The second test for the `serve/` directory is simpler but justified since static asset serving is discovered at runtime via `express.static`, not through `require()`. Both tests verify behavior (will the package work when published?) rather than implementation details.

One observation: `collectLocalRequires` uses regex-based static analysis (`/require\(["'](\.[^"']+)["']\)/g`) rather than actually executing the modules. This is the right trade-off — it avoids needing a running server — but it will miss dynamic requires like `require(variable)`. Not a problem today since the codebase doesn't use that pattern, but worth noting in case it ever does.

**Principle 6 (Refactoring is not optional):** The structural refactor here is clean. Moving constants from `shared/` into `server/` with re-export shims in `shared/` is the right direction — it solves the npm packaging constraint without breaking the feedback-layer build. The re-export shims are a named compromise: they exist because of a real constraint (npm can't follow parent-directory references), they're commented, and they're minimal (1 line each).

**Principle 12 (Do it right the first time):** The Dockerfile update correctly removes the two separate `COPY` lines and replaces them with one that matches the new layout. The eslint config, build script, and static path all got updated to match. No stale references left behind.

**Principle 3 (Red-green-refactor):** The commit history shows the fix and the test arriving together in one commit, with a follow-up for eslint config. Given the nature of the change (file moves + package.json updates), this is acceptable — the "test" here is really a regression guard for a packaging problem that was discovered manually.

No issues found. Ship it.

PR #262, opened 2026-03-17
feat: add annotation exports (JSON, CSV, PDF) (#61)

### LINKED ISSUE #61: Annotation exports (JSON, CSV, PDF)
## Jobs to Be Done

- As a **document author**, I want to export all annotations on my document as JSON, CSV, or PDF, so that I can archive feedback, share it with stakeholders, or feed it into other tools.
- As an **AI agent developer**, I want a bulk export format, so that I can ingest all feedback in a single request rather than paginating through the API.

## Technical Proposal

Add export endpoints to the server: `GET /documents/:id/export?format=json|csv|pdf`. JSON export is trivial — serialize the comments array with document metadata. CSV export flattens each comment into a row (columns: quote, body, author, status, created_at, parent_id). PDF export requires a rendering library — use something lightweight like `pdfkit` to generate a document listing each annotation with its quoted text and comment body. The client adds an "Export" button to the sidebar header with a format dropdown. Exports are generated server-side and returned as downloadable files with appropriate `Content-Disposition` headers.

cass-clearly:
## Summary
- Adds export endpoint to server: `GET /documents/:id/export?format=json|csv|pdf`
- JSON export serializes comments array with document metadata
- CSV export flattens each comment into a row (quote, body, author, status, created_at, parent_id)
- PDF export uses pdfkit to generate a document listing each annotation with quoted text
- Client adds an "Export" button to sidebar header with format dropdown
- Exports are generated server-side and returned as downloadable files with Content-Disposition headers

Also fixes a pre-commit hook bug (wrong path to feedback-layer.js).

Closes #61

## Test plan
- [x] Server tests pass including new export endpoint tests
- [x] Client tests pass
- [ ] Manual test: export annotations in JSON format
- [ ] Manual test: export annotations in CSV format  
- [ ] Manual test: export annotations in PDF format


cass-clearly (commented):
## Review: feat: annotation exports (JSON, CSV, PDF)

### What's Good
- Three export formats implemented cleanly: JSON (full data), CSV (flat, good for spreadsheets), PDF (human-readable)
- Proper `Content-Disposition: attachment` headers with descriptive filenames
- pdfkit is the right lightweight choice for PDF generation without a browser rendering dependency  
- Export button in sidebar header with format dropdown is a clean UX pattern
- 6 new server tests cover all formats + error cases (missing format, invalid format, missing document)
- 145 server + 91 client tests all pass

1. **CSV escaping**: Verify that quote/body fields with commas, newlines, or double-quotes are properly escaped (RFC 4180). If comments can contain multiline text, this is important.
2. **Large documents**: For a document with thousands of comments, the PDF export could be slow. Acceptable for v1 — streaming could be a follow-up if needed.
3. **Authentication**: Export endpoints are unprotected (consistent with the rest of the API). If rate limiting from #186 is applied globally, exports are covered.

Strong implementation that closes a real user need. Approve.

cass-clearly:
## 🔪 The Minimalist — Round 1 Review

**Verdict: REQUEST CHANGES**

---

### What this does

Export annotations in JSON, CSV, or PDF. Three formats. One new endpoint. One new npm dependency (pdfkit). UI dropdown in the sidebar.

---

### Issues

**1. PDF export belongs in the client, not the server.**

JSON and CSV are trivial data transformations — they belong on the server. PDF generation is different. You pulled in `pdfkit` (which drags in `fontkit`, `brotli`, `@swc/helpers`, `@noble/ciphers`, `@noble/hashes`, `dfa`, `clone`, and more — see the lock file). This is a large dependency tree for a feature that could be implemented in the browser with `window.print()` and a print stylesheet, or by constructing the download client-side from the JSON export.

The browser already has the data. The server doesn't need to render it into a document. Principle 14: just-in-time over just-in-case. Principle 1: delete first.

**Fix:** Delete PDF from the server. If PDF is genuinely required: implement it client-side, or generate it from the JSON response. Do not add a server-side PDF renderer for a sidebar export button.

**2. `escape` helper is defined inline inside a `map` callback.**

In the CSV handler:
```js
const csvRows = commentsWithReactions.map((c) => {
  const escape = (v) => { ... };
  return [...].join(",");
});
```

This function is re-created on every iteration. Either hoist it or inline the two-line body. This is not a style concern — it's clarity and principle 4.

**Fix:** Extract `escape` above the `map` call.

**3. `handleExport` in the client has a fragile document ID fallback.**

```js
const documentId = _docId || (_comments.length > 0 ? _comments[0].document : null);
```

This was already the pattern used elsewhere in this codebase (e.g., WebSocket init). That doesn't make it good here. If you're adding a new code path, don't copy the hack — fix it. The document ID should be reliably set from config at boot, not inferred from comment data.

This isn't a blocker but it's worth naming. The fragility is pre-existing; don't spread it.

**4. Export endpoint is unauthenticated and undocumented.**

Every comment, every author, every reaction — fully accessible to anyone who knows (or guesses) a document ID. This project has no auth today, so that may be acceptable. But the export endpoint makes the exposure significantly larger since it returns the entire annotation corpus in one request. If there's no auth plan, at minimum document this in a comment on the endpoint or in CLAUDE.md. Principle 12: if you're cutting a corner, name it.

**5. Tests create a fresh document per test case in the export suite — that's correct but the pattern is verbose.**

Five test cases, each does a full document + comment setup. No shared setup using `before`/`beforeEach`. Look at the existing test patterns in this file — they use before hooks. Align with the existing pattern.

---

### What's fine

- JSON and CSV export logic: clean, correct, minimal.
- Tests cover the happy paths and error cases.
- No functional bugs found in the server logic (tests pass: 138 passing, 1 pre-existing unrelated failure).
- Client-side dropdown implementation is straightforward.

---

### Required changes

1. Delete PDF from the server. Either remove it or move generation client-side.
2. Hoist the `escape` function out of the `map` callback.
3. Use `before`/`beforeEach` for shared test setup in the export suite.

cass-clearly:
## 🏛️ The Architect — Round 1 Review

**Verdict: APPROVE**

---

### Summary

This PR is coherent, scoped correctly, and ships a complete feature. Server-side export (JSON/CSV/PDF), client UI (dropdown in sidebar header), and tests all land together. The pre-commit hook fix is a legitimate bonus. Tests pass — 139 server, 91 client, all green.

---

### What holds together

**The architecture is right.** Server-side generation is the correct call for exports at this fidelity level. PDF rendering in the browser would be fragile, client-side CSV generation would miss reactions, and streaming blobs via Content-Disposition is the right HTTP primitive here. The decisions made here are all defensible.

**Scope discipline is good.** This PR does one thing: annotation exports in three formats. The pre-commit hook fix shares a natural boundary (it was broken, it's touched anyway). Nothing extraneous crept in.

**The implementation is proportionate.** The server handler is ~120 lines, the client adds one button and one dropdown. PDFKit is a reasonable dependency for what's needed. The dep tree it pulls (fontkit, brotli, etc.) is heavier than I'd like but it's the right tradeoff unless we want to either hand-roll PDF generation or add a microservice — neither is worth it.

**The client-side document ID resolution in handleExport deserves mention:**
```js
const documentId = _docId || (_comments.length > 0 ? _comments[0].document : null);
```
This works, but it reveals that `_docId` may not always be set when comments exist. It's not a bug for the happy path but it's a latent fragility. File it as a follow-up, don't block on it.

---

### What's missing

**The export endpoint is absent from openapi.js.** This is a real omission — not a cosmetic one. The OpenAPI spec is how consumers discover what's available and what the contract is. Not a blocker for shipping, but it should be filed immediately as a follow-up issue. API stability is a principle here (principle 8), and undocumented endpoints don't get that protection.

**Manual test plan items are unchecked.** Three manual test items are unchecked in the PR description. The automated tests cover the contract well (including the %PDF- header check), but the manual items exist for a reason — user-visible behavior. This is fine to ship with, just noting it.

**CSV doesn't include reactions.** The `reactions` field on comments is populated but silently dropped in the CSV serialization. JSON exports get reactions. PDF exports ignore them too. Whether this matters depends on how reactions are used downstream, but it's an inconsistency worth noting. Not blocking.

---

### Risk assessment

**Actual risk: Low.** The endpoint is additive (new route, no changes to existing routes). The client change adds a button that calls a new endpoint — failure is a toast error, nothing breaks. PDFKit is a well-established library. The dependency additions are pure additions to the server package.

**The one thing I'd watch:** `doc.pipe(res)` in the PDF path — if an error occurs mid-stream after headers are sent, the client will receive a partial PDF with a 200 status. This is a known limitation of streaming responses. It won't cause data corruption but it could produce a broken download. Wrapping the PDFDocument generation in more defensive error handling (or at minimum logging the error event on the doc stream) would be prudent. File as follow-up.

---

### Follow-up issues to file

1. Document the export endpoint in `openapi.js`
2. Add reactions to CSV export (or explicitly document that CSV does not include reactions)
3. Add error handling for mid-stream PDF failures
4. Clarify `_docId` initialization guarantees (or document the fallback as intentional)

---

**This is done enough to ship.** The core feature works, tests pass, the architecture is right, and the gaps are all follow-up quality. Approve.

cass-clearly:
## 🔨 The Craftsperson — Round 1 Review

**Verdict: REQUEST CHANGES**

Tests pass (145/145 ✅). The feature works. But the refactor step was skipped, and there are code quality issues that will compound over time. Here's what needs to change before this merges.

---

### 🔴 Critical: Massive route handler violates Single Responsibility

**File:** `server/index.js`, the `/documents/:id/export` route

This handler is doing six different things in one function: parsing format, fetching document, fetching comments, fetching reactions, building CSV, building JSON, building PDF. It's 80+ lines in a single `asyncHandler` callback.

Apply the **Extract Function** refactoring:

```js
// server/export.js (new file)
function buildJsonExport(document, commentsWithReactions) { ... }
function buildCsvExport(commentsWithReactions) { ... }
async function streamPdfExport(document, commentsWithReactions, res) { ... }
```

Then the route becomes:

```js
app.get("/documents/:id/export", asyncHandler(async (req, res) => {
  const { format } = req.query;
  if (!isValidFormat(format)) {
    return res.status(400).json(errorResponse("format must be json, csv, or pdf"));
  }
  const document = await fetchDocumentOrFail(req.params.id);
  if (!document) return res.status(404).json(errorResponse("Document not found"));
  const comments = await fetchCommentsWithReactions(req.params.id);
  return formatHandlers[format](document, comments, res, req.params.id);
}));
```

This route handler currently cannot be tested without spinning up the full Express + Postgres stack. The rendering logic (CSV escaping, PDF tree building, JSON structure) should be unit-testable in isolation.

---

### 🔴 Code smell: Inline function defined inside a `.map()` callback

**File:** `server/index.js`, CSV export block

```js
const csvRows = commentsWithReactions
  .map((c) => {
    const escape = (v) => {   // ← redefined on every iteration
      if (v === null || v === undefined) return "";
      const str = String(v).replace(/"/g, '""');
      return `"${str}"`;
    };
    return [...].join(",");
  })
```

The `escape` function is redefined on every call to `.map()`. Extract it:

```js
function escapeCsvValue(v) {
  if (v === null || v === undefined) return "";
  return `"${String(v).replace(/"/g, '""')}"`;
}
```

---

### 🔴 PDF status label bug — wrong fallback value

**File:** `server/index.js`, PDF rendering

```js
doc.text(`— ${comment.author} | ${comment.status || "reply"} | ${comment.created_at}`);
```

A top-level comment with `status: null` renders as `"reply"` — which is wrong. This is a top-level comment, not a reply. The fallback should be `"open"` or `"pending"`. No test covers this case. The existing PDF test creates a comment with status `"open"`, so this slip went undetected.

Fix:
```js
doc.text(`— ${comment.author} | ${comment.status ?? "open"} | ${comment.created_at}`);
```

And add a test:
```js
it("PDF labels null-status comments correctly", async () => { ... });
```

---

### 🟡 Test smell: Duplicate setup across all six export tests

Every export test creates a document and adds a comment with the same boilerplate. That's **Duplication of Knowledge** — if the setup changes, six tests need updating.

Extract a test helper:

```js
async function createDocumentWithComment(overrides = {}) {
  const docRes = await fetch(`${BASE}/documents`, {
    method: "POST",
    headers: { "Content-Type": "application/json" },
    body: JSON.stringify({ uri: "https://example.com/page" }),
  });
  const doc = await docRes.json();
  await fetch(`${BASE}/comments`, {
    method: "POST",
    headers: { "Content-Type": "application/json" },
    body: JSON.stringify({ document: doc.id, quote: "text", body: "hello", author: "me", ...overrides }),
  });
  return doc;
}
```

Then each test becomes three lines instead of fifteen.

---

### 🟡 No test for reactions in JSON export

The handler fetches reactions and attaches them to comments, but no test verifies they appear in the exported JSON. A regression in `fetchReactionsForComments` would silently drop reactions from exports. Add:

```js
it("includes reactions in JSON export", async () => {
  const doc = await createDocumentWithComment();
  // POST a reaction to the comment...
  const res = await fetch(`${BASE}/documents/${doc.id}/export?format=json`);
  const json = await res.json();
  assert.ok(Array.isArray(json.comments[0].reactions));
  assert.equal(json.comments[0].reactions.length, 1);
});
```

---

### 🟡 Integration-only test coverage for rendering logic

CSV escaping, JSON structure, and PDF tree building are tested only through the HTTP layer. This means you can't test edge cases (commas in quotes, unicode in author names, empty comment bodies) without hitting the database.

Once the rendering functions are extracted (see critical item #1), add unit tests:

```js
describe("buildCsvExport", () => {
  it("escapes double quotes in values", () => { ... });
  it("handles null status", () => { ... });
  it("handles null parent_id", () => { ... });
  it("includes reactions column if present", () => { ... });
});
```

---

### 🟢 What's done well

- **Validation is tight.** The allowlist check on `format` before any DB queries is correct — fail fast.
- **Tests cover the HTTP contract.** Status codes, content-type headers, content-disposition headers, PDF magic bytes (`%PDF-`) — all verified.
- **Frontend integration is clean.** The `onExport` callback pattern follows the existing sidebar callback convention correctly. The document ID fallback logic (`_docId || _comments[0].document`) is sensible.
- **No breaking API changes.** New endpoint only, additive. Existing consumers unaffected.
- **Error boundaries.** The client-side `handleExport` catches errors and shows toasts. Server errors propagate correctly.

---

### Summary

The green path works and is tested. The refactor step was not done. The export route is a God Function doing too much in one place. The CSV helper has a smell, there's a latent bug in the PDF status label, and the tests duplicate setup that will be painful to maintain. Six test cases doing 15 lines of setup each when one helper would do it is a clear signal the refactor step was skipped.

**Make it right: extract the format renderers into a separate module, fix the PDF status fallback, add a reactions integration test, and consolidate the test setup into a shared helper.**

cass-clearly:
## 🛡️ The Steward — Round 1 Review

**Verdict: REQUEST CHANGES**

Tests pass (145/145, 94.34% line coverage on index.js). The implementation is solid in most respects. Three issues need to be addressed before this merges.

---

### 🔴 Issue 1: OpenAPI spec not updated (API stability — Principle 8)

`GET /documents/:id/export?format=json|csv|pdf` is a new public endpoint. The OpenAPI spec in `server/openapi.js` has not been updated to document it.

This matters because the OpenAPI spec is the contract. Consumers relying on generated clients or the `/openapi.json` endpoint will have no idea this endpoint exists. If something must break for existing consumers, there's a deprecation path — but for new endpoints, there's no excuse for skipping documentation. The spec isn't optional; it's how we tell the world what the API does.

Add the path entry for `/documents/{id}/export` with the `format` query parameter (enum: json, csv, pdf), all three response types (application/json, text/csv, application/pdf), and the 400/404 error cases.

---

### 🔴 Issue 2: README not updated (consumer experience — Principle 8)

The README documents every other endpoint. This one isn't there. A developer reading the README after this ships will not find export. A developer using the CLI won't know this exists.

Add a section for the export endpoint. Show the curl examples. Document what each format returns.

---

### 🟡 Issue 3: PDF generation has no memory safeguard for large documents

`PDFDocument` buffers in memory during generation. For a document with hundreds of annotations with long quotes and bodies, this will consume significant memory and the response will hang until generation completes — or the process will OOM.

This won't be a problem in early usage, but it's a failure mode worth naming. The current architecture is fine for MVP, but the comment in the code should acknowledge this:

```js
// NOTE: PDFDocument buffers in memory. For very large annotation sets, consider
// streaming or paginating. Current approach is acceptable for expected document sizes.
```

Name the corner cut. Don't let it become invisible debt (Principle 12).

---

### ✅ What's done well

- **Input validation is correct.** 400 for missing/invalid format, 404 for missing document — exactly right.
- **Consistent error response shape.** Uses `errorResponse()` like every other endpoint.
- **Test coverage is thorough.** All six cases tested: JSON, CSV, PDF, missing format, invalid format, missing document. PDF test validates the `%PDF-` magic bytes — that's the right level of rigor.
- **Client-side download implementation is clean.** The `URL.createObjectURL` → click → revoke pattern is the correct browser-native approach.
- **Pre-commit hook path fix is correct and needed.**
- **No new query patterns introduced.** Export reuses `fetchReactionsForComments` and follows the same DB access patterns as the rest of the codebase.
- **No auth changes.** Export is read-only and follows the same access model as other document reads.

---

### Summary

The feature works. The implementation is production-appropriate. What's missing is the paper trail: OpenAPI spec and README are the contract this project maintains with its consumers. Both need to reflect this endpoint before it ships.

Fix the spec and docs, name the PDF memory corner cut, then this is good to go.

cass-clearly:
## 📣 The Marketing Guru — Round 1 Review

**Verdict: REQUEST CHANGES**

---

### What this PR does

Adds `GET /documents/:id/export?format=json|csv|pdf` to the server, plus an Export button + format dropdown in the feedback-layer sidebar. Tests pass — 0 failures across all suites. The implementation is clean and the test coverage is solid (JSON, CSV, PDF, 400 for missing/invalid format, 404 for missing document).

---

### 🔴 CHANGELOG.md — Not documented

The `[Unreleased]` section has **no entry for this feature**. This is a required check and it fails.

Every shipped feature earns a CHANGELOG entry. Export is a major user-facing feature — arguably the most immediately marketable thing in recent memory — and it's invisible to anyone reading the changelog.

Required entry, something like:

```markdown
### Added
- Annotation export: download all comments for a document as JSON, CSV, or PDF
  via `GET /documents/:id/export?format=json|csv|pdf`. JSON includes document
  metadata; CSV flattens one comment per row; PDF generates a readable report.
  Export button added to sidebar header.
```

The description should sell the capability, not just describe the implementation.

---

### 🔴 docs/api.md — Export endpoint not documented

The API reference has no entry for `GET /documents/:id/export`. This is a public endpoint. It needs:

- Method, path, description
- Query parameters (`format`: required, `json` | `csv` | `pdf`)
- Response headers (`Content-Type`, `Content-Disposition`)
- Response body (for JSON: schema; for CSV/PDF: note on binary/text response)
- Error cases (400 for missing/invalid format, 404 for missing document)
- A copy-pasteable `curl` example for each format

Agents reading our API docs to help integrators will have no idea this endpoint exists.

---

### 🟡 README — Export not surfaced

The README doesn't mention export. This is lower priority than the API docs, but if we're positioning Remarq as the alternative to commenting in Google Docs, **data portability is a selling point**. Google Docs lets you export your *document* — Remarq lets you export your *annotations*. That's a story worth telling.

Consider adding a one-liner under features or the quick start: "Export annotations as JSON, CSV, or PDF."

---

### 🟢 Marketing opportunity — this is genuinely good

Export is the feature that answers "what do I do with all these annotations once the review is done?" Google Docs has comments. You can't export them as structured data. You can't pipe them into a spreadsheet, feed them into a report, or archive them as JSON. Remarq can.

This is a concrete, demonstrable advantage. A changelog entry alone isn't enough — this deserves a sentence on the landing page under the feature highlights, and it's strong material for a future release post. Flag it for the CMO/content pipeline.

---

### ✅ Implementation quality

- Tests are well-structured and cover the right cases
- 400 on missing format, 400 on invalid format, 404 on missing document — all handled correctly
- Content-Disposition headers included — downloadable out of the box
- PDF uses pdfkit — reasonable choice, no heavy dependencies

No code changes requested. The implementation is fine. The docs are not.

---

### Summary

| Check | Status |
|---|---|
| CHANGELOG.md under [Unreleased] | ❌ Missing |
| docs/api.md updated | ❌ Missing |
| README mentions export | ⚠️ Missing (minor) |
| Tests pass | ✅ |
| Marketing opportunity identified | ✅ |

Blocking on CHANGELOG and API docs. Both are quick — this PR is one good commit away from being mergeable.

cass-clearly (commented):
**The Minimalist (Round 1):**

This PR adds annotation export in three formats (JSON, CSV, PDF). The feature is grounded in a real issue (#61), the implementation is appropriately lean, and I largely approve. A few notes worth flagging:

**What I like:**
- The three export modules are dead simple — `renderJson` is 5 lines, `renderCsv` is ~20, `renderPdf` is ~40. No unnecessary abstraction. No base class, no factory, no strategy pattern. Just functions.
- No new middleware, no new service layer. The route handler does exactly what it needs to.
- The `escape()` function in `export-csv.js` is small, exported, and tested. That's the right level of decomposition.

**Questions / concerns:**

1. **Three separate module files for three small functions.** `export-json.js` is 12 lines. Could these three live in a single `export.js` file? The overhead of navigating three files feels higher than the benefit of separating them. I'm not blocking on this — it's a taste call — but it's worth naming.

2. **pdfkit dependency.** PDFKit adds ~1MB to the server bundle. That's non-trivial. Is PDF export a near-term user need, or is this speculative? If most users will only use JSON or CSV for the next 6 months, shipping the PDFKit dependency today is just-in-case, not just-in-time. No strong objection, but I'd like to see evidence it was asked for, not assumed.

3. **The client export dropdown is custom-built.** A `<select>` would be functionally equivalent and ~0 lines of CSS. The custom dropdown requires ~30 lines of CSS (`fb-export-container`, `fb-export-btn`, `fb-export-dropdown`, `fb-export-option`) and event wiring to close-on-outside-click. This is the kind of complexity that grows. Worth questioning whether the UX benefit justifies it.

**Not blocking on any of these.** The core feature is appropriately minimal, well-tested, and ships a real user need. These are refinements, not rejections.

cass-clearly (commented):
**The Craftsperson (Round 1):**

This is solid work overall. Tests are well-structured and cover the right surface area. A few things worth addressing:

**Test quality — good:**
- Unit tests for each export module (`export-json`, `export-csv`, `export-pdf`) are written at the right level — they test behavior, not implementation. I can refactor internals and these tests won't break.
- The CSV escape tests are exemplary: they test the edge cases that matter (`null`, `undefined`, internal double quotes).
- Integration tests cover 400/404 error paths, not just the happy path.
- The `makeComment()` helper with overrides is clean test factory pattern.

**Naming and readability:**
- `renderJson`, `renderCsv`, `renderPdf` — consistent, intent-revealing. Good.
- `commentsWithReactions` as a parameter name is descriptive. Good.
- The route handler is longer than it needs to be because it has three format branches that are structurally identical (fetch doc → fetch comments → fetch reactions → render). This is a mild duplication smell, but the three branches are simple enough that inlining keeps it readable. Not requesting a refactor.

**One real concern — the `null` status test is prescriptive without enforcement:**

The test `handles null status (renders as 'open', not 'reply')` in the CSV suite asserts:
```js
assert.equal(cols[3], ); // status column is empty
```
But the comment says "null status should render as empty string, not 'null'". The `escape()` function correctly handles null → empty string. However, the PDF test has a similar assertion:
```js
assert.ok(!pdfText.includes("| reply |"), "PDF should not render 'reply' as status...");
```
This test is inspecting raw PDF bytes for the string "| reply |" — which doesn't appear in `renderPdf` at all. The PDF renders `comment.status ?? "open"`, not with pipe delimiters. This test **will pass vacuously** — it's asserting the absence of something that was never generated. It doesn't protect against regressions in null status handling. The assertion should be `pdfText.includes("open")` to actually verify the fallback.

**Error handling:**
The `renderPdf` function correctly wires `doc.on("error", reject)` — good. The route handler uses `asyncHandler` which propagates errors correctly. No gap here.

**One request: fix the vacuous PDF null-status test to assert presence of "open" rather than absence of a string that can't appear.**

cass-clearly (commented):
**The Steward (Round 1):**

This PR adds a well-designed export endpoint with appropriate production hygiene. A few observations from the operational and consumer trust perspective:

**API contract — good:**
- The endpoint follows the existing resource-based pattern: `GET /documents/:id/export?format=...`. Consistent with the rest of the surface.
- 400 for missing/invalid format, 404 for missing document — correct error semantics.
- `Content-Disposition: attachment` with a sensible filename (`annotations-{id}.{format}`) — consumers get a usable file without extra work.
- `Content-Type` headers are correct for all three formats.
- OpenAPI spec is updated (verified in test: `assert.ok(json.paths["/documents/{id}/export"])`) — consumers can discover this without reading source.

**Documentation:**
- README update adds the endpoint with curl examples. Spot-checked and the examples look accurate.
- The API docs (`docs/api.md`) were updated. Good.

**Operational concerns (minor):**

1. **Memory for PDF export.** `export-pdf.js` buffers the entire PDF in memory before sending. The code comments this explicitly with a NOTE. That's the right behavior for MVP scale. No action needed — the comment is the documentation.

2. **No timeout on PDF generation.** If pdfkit hangs for any reason (unlikely but possible with malformed data), the request will hang until the client gives up. The `asyncHandler` doesn't add a timeout. For a v1 MVP this is acceptable, but worth filing as a follow-up issue.

3. **Content-Type header for JSON.** The route sets `res.setHeader("Content-Type", "application/json")` and then calls `res.json()`, which also sets Content-Type. No harm in this, but it's slightly redundant. Express's `res.json()` already sets the correct content type. Not requesting a change — just noting the redundancy.

4. **`charset=utf-8` in content-type.** The integration test asserts `res.headers.get("content-type") === "application/json; charset=utf-8"` — this works because `res.json()` appends charset automatically. The CSV endpoint asserts `includes("text/csv")` without charset — which is a small gap if consumers depend on charset being present. Low priority.

5. **Changelog.** Updated. Good.

**Verdict:** This is production-ready for MVP scale. The PDF memory note is documented, the API surface is clean, and the error paths are correct. No blocking concerns.

cass-clearly (commented):
**The Architect (Round 1):**

All three council members reviewed. Here's my synthesis:

**Assessment: Approve with one fix requested.**

The core feature is well-scoped, correctly implemented, and production-ready. Three small export modules, a clean route handler, full test coverage (unit + integration), updated docs and OpenAPI spec, a pre-commit hook bugfix bundled in. This is how new features should be built.

**Council signal vs. noise:**

The Minimalist raised three points:
1. *Three files vs one file* — noise. Separate files per format is the right call if these grow. No change needed.
2. *pdfkit dependency* — valid question. Issue #61 asked for PDF export; it's not speculative. The dependency is justified. No change needed.
3. *Custom dropdown vs select* — mild signal. A custom dropdown is heavier than a `<select>`, but it's consistent with the existing custom UI patterns in the sidebar. If we want to normalize toward native elements that's a separate PR. No change needed here.

The Steward raised operational notes that are all non-blocking and largely already acknowledged in the code comments. The redundant Content-Type header on the JSON endpoint is a 1-line cleanup worth doing, but not a blocker.

The Craftsperson identified one **real bug**: the PDF null-status test asserts absence of `"| reply |"` — a string that never appears in the PDF output. This test passes vacuously and provides no regression coverage. **This should be fixed before merge.** Change the assertion to verify that the fallback `"open"` appears in the rendered PDF text.

**Specific action requested:**

In `server/test.mjs`, the test `"handles comments with null status (renders as 'open', not 'reply')"`:


**Cleanup (non-blocking, do it anyway):** Remove the redundant `res.setHeader("Content-Type", "application/json")` line before `res.json()` in the export route handler. `res.json()` sets it automatically.

Once those two changes land and tests pass, this is approved.

cass-clearly (commented):
**The Architect (Round 1):**

All three council members reviewed. Here's my synthesis:

**Assessment: Approve with one fix requested.**

The core feature is well-scoped, correctly implemented, and production-ready. Three small export modules, a clean route handler, full test coverage (unit + integration), updated docs and OpenAPI spec, a pre-commit hook bugfix bundled in. This is how new features should be built.

**Council signal vs. noise:**

The Minimalist raised three points:
1. *Three files vs one file* — noise. Separate files per format is the right call if these grow. No change needed.
2. *pdfkit dependency* — valid question. Issue #61 asked for PDF export; it's not speculative. The dependency is justified. No change needed.
3. *Custom dropdown vs select* — mild signal. A custom dropdown is heavier than a select, but it's consistent with the existing custom UI patterns in the sidebar. If we want to normalize toward native elements that's a separate PR. No change needed here.

The Steward raised operational notes that are all non-blocking and largely already acknowledged in the code comments. The redundant Content-Type header on the JSON endpoint is a 1-line cleanup worth doing, but not a blocker.

The Craftsperson identified one **real bug**: the PDF null-status test asserts absence of `"| reply |"` — a string that never appears in the PDF output. This test passes vacuously and provides no regression coverage. **This should be fixed before merge.** Change the assertion to verify that the fallback `"open"` appears in the rendered PDF text.

**Specific action requested:**

In `server/test.mjs`, the test `"handles comments with null status (renders as 'open', not 'reply')"`:

Current (vacuous):
```
assert.ok(!pdfText.includes("| reply |"), "PDF should not render 'reply' as status...");
```

Fix to:
```
assert.ok(pdfText.includes("open"), "PDF should render 'open' as fallback for null status");
```

**Cleanup (non-blocking but do it anyway):** Remove the redundant `res.setHeader("Content-Type", "application/json")` line before `res.json()` in the export route handler. `res.json()` sets it automatically.

Once those two changes land and tests pass, this is approved.

cass-clearly (commented):
**Council Decision (Round 2): APPROVED — Ready to merge.**

The author addressed all council feedback proactively. Commit 6191774 ('fix: real null-status PDF assertion + remove redundant Content-Type header') resolves both items The Architect requested:

1. **Vacuous PDF test fixed** — replaced the absence assertion with a proper zlib decompression approach that inflates the PDFKit FlateDecode stream and extracts hex-encoded TJ text segments, then asserts '| open |' appears in the rendered text. This test now provides real regression coverage.

2. **Redundant Content-Type header removed** — the redundant res.setHeader before res.json() is gone.

All 157 tests pass including the new null-status PDF test.

**The council's collective verdict:**
- The Minimalist: no blocking concerns, minor preferences noted for future consideration
- The Craftsperson: test quality is good, the one real issue is now fixed
- The Steward: production-ready for MVP scale, operational notes documented
- The Architect: coherent, well-scoped, ship it

This PR is approved for merge.

